PR #3, opened 2021-12-17
Add notebooks

ackagel:
adds notebooks, conda environment file, and installation instructions


ngreenwald:
I think this one has enough in it for one PR, take a look through and let me know what you think @ackagel

creed/qc_comp.py
```diff
@@ -0,0 +1,505 @@
...
+from shutil import rmtree
+from skimage.io import imsave
+
+from creed.mibitracker_utils import MibiTrackerError
+from creed.mibitracker_utils import MibiRequests
+
+import ark.settings as settings
+import ark.utils.io_utils as io_utils
+import ark.utils.load_utils as load_utils
+import ark.utils.misc_utils as misc_utils
```
ackagel, line 18: I feel like we could avoid these ark imports for a future version.  Either via reimplementing the requisite tools in the creed repo, or consolidating our io related code into another separate repo/project (which sounds silly, but in almost every mibi-related project, I usually end up installing the entirety of ark for these io/loading utils anyways)

creed/qc_comp_test.py
```diff
@@ -0,0 +1,331 @@
...
+    (['fov0', 'fov1'], None, False),
+    (['fov0', 'fov1'], None, True),
+    (None, ['chan0', 'chan1'], False),
+    (None, ['chan0', 'chan1'], True),
+    (['fov0', 'fov1'], ['chan0', 'chan1'], False),
+    (['fov0', 'fov1'], ['chan0', 'chan1'], True)
+]
+
+MIBITRACKER_EMAIL = 'qc.mibi@gmail.com'
+MIBITRACKER_PASSWORD = 'The_MIBI_Is_Down_Again1!?'
```
ackagel, line 43: sigh

ackagel: i think the long term solution for this type of problem is to just make this an environment variable.  that can be set in travis pretty easily I think, and as for local testing, we just require you set that env variable before testing.

ackagel (commented):
backend stuff looks fine.  there's definitely room for a refactor but nothing urgent.  Still need to look at the notebooks on nbreview, then I'll approve

templates/example_qc_metric_eval.ipynb
```diff
@@ -0,0 +1,425 @@
+{
```
ackagel, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    password = 'The_MIBI_Is_Down_Again1!?'
</pre><p>sigh</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/creed-helper/pull/3/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

ackagel (commented):
kk looks good. there's some rocky looking stuff (mostly from my notebooks) that should be refactored at some point, but it's fine for now

creed/qc_comp.py
ngreenwald: Agreed. This would be a nice starter project for our new hires

creed/qc_comp_test.py
ngreenwald: Yup. We can also use notebook injection so the password isn't visible here, it could just be a placeholder.

PR #15, opened 2022-01-14
Add Travis and pytest support

alex-l-kong:
Add configuration files to submit repo to Travis on commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
-dist: xenial
+jobs:
+  include:
+    -
+      os: linux
+      dist: xenial
+    -
+      os: windows
 
 git:
   depth: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ jobs:
     -
       os: linux
       dist: xenial
-    -
-      os: windows
 
 git:
   depth: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ cache: pip
 
 script:
   - python -m pip install --editable .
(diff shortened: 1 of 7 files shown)


requirements-test.txt
```diff
@@ -0,0 +1,7 @@
+coveralls>=1.8.2,<2.0.0
+pytest<6.0.0
+pytest-cov<3.0.0
+pytest-mock<4.0.0
+pytest-pycodestyle<3.0.0
+pytest-asyncio>=0.15.1
```
ackagel, line 6: spoiler alert, get `pytest-asyncio` outa here (it's evil and will crash pytest on travis)

.travis.yml
```diff
@@ -0,0 +1,25 @@
+dist: xenial
```
ackagel, line 1: we might want to look into testing this on a windows dist if that's an option

alex-l-kong: Yeah I added multiple `jobs` to the `.travis.yml` file to include two `os` args: `xenial` and `windows`. We can probably remove `xenial` since I don't think anyone will be running these notebooks on non-Windows machines.

alex-l-kong: UPDATE: nvm, `python` is unsupported on the Windows dist, reverting back to `dist: xenial`.

alex-l-kong:
All should be good now, one thing to note is that due to the nature of `numpy` and their recent releases, we need to temporarily set the `numpy` version to `1.22` in `requirements.txt` for it to play nicely with our notebooks and dependent GitHub repos (`ark-analysis` and `mibi-bin-tools`).

PR #17, opened 2022-01-15
Add automatic detector sweep ID

ngreenwald:
Detector sweeps generate a stereotyped naming scheme of Detector_ voltage _ date _ time, e.g. `Detector_2380v_2022-01-14_14-06-36.` We often want to get all of the files associated with a particular sweep. If only one sweep was performed that day, this is easy, but if multiple sweeps were performed there isn't always a unique prefix that will grab all of the relevant files. This PR adds a function to identify those sweeps. 

Tests are left as an exercise for ~~the reader~~ whoever has time later.


creed/misc.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def find_detector_sweep_folders(data_dir, first_fov, last_fov, sweep_step=25):
+    """Generates the names of all folders from a single detector sweep
+
+    Args:
+        data_dir: directory where runs are saved
+        first_fov: the name of the first FOV from the detector sweep
+        last_fov: the name of the last FOV from the detector sweep
+        sweep_step: the voltage increase between FOVs in the detector sweep
```
ackagel, line 27: any downside of finding all fovs with times in between first and last, and then filtering by voltage intervals afterwards?

creed/misc.py
```diff
@@ -0,0 +1,67 @@
+from ark.utils import io_utils
+
+
+def extract_sweep_parameters(sweep_name):
```
ackagel, line 4: kind of a future direction here, but pivoting towards using python dataclasses for could help to 'better document' + 'unify usage of' these simple, consistent data structures.  In this case, something like:
```
@dataclass
class VoltageSweepFile:
    voltage: int
    date: str
    time: str


def parse_sweep_name(sweep_name: str) -> VoltageSweepFile:
    ...
    return VoltageSweepFile(voltage, date, time)
```

would allow stuff like:
```
first_sweep = parse_sweep_name(first_fov)
last_sweep = parse_sweep_name(last_fov)

if first_sweep.date != last_sweep.date:
    ...
```

maybe a bit 'over-organizing' in this case, but I think getting into the habit of using these could be good.

ackagel (commented):
looks good for the most part.  Also, maybe change the filename from `misc.py` to `detector_sweep.py` maybe?

ngreenwald:
Good points, I'll add them in.

alex-l-kong (commented):
Nothing on top of Adam to add. Looks good to me.

ngreenwald:
Will re-request review once I or someone else fills out the tests

alex-l-kong (commented):
Just one minor comment about adding a test.

ackagel (commented):
looks good, no complaints from me

alex-l-kong (approved):
Looks good!

ackagel (approved):

PR #27, opened 2022-01-23
Add Rosetta matrix functionality

ngreenwald:

Adds in key functions to run the rosetta matrix on mibi data. 

The user provides a csv file with the mapping of each mass to an antibody target, as well as a compensation matrix with coefficients for all masses. The code will then compensate the data using the provided matrix, and save the outputs. There is also a function to compare the outputs from multiple rosetta runs in a single tiled image, and the backbone of the function that will be used to generate those different directory structures.


creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Args:
+        json_path: path to json file
+        comp_mat_path: path to comp matrix
+
+    returns:
+        pd.DataTable: matrix with sources channels as rows and target channels as columns"""
+
+    with open(json_path) as f:
+        data = json.load(f)['Data']
```
alex-l-kong, line 26: Might be a good idea to validate `json_path` and `comp_mat_path` exists.

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+        gaus_rad: radius for blurring image data. Passing 0 will result in no blurring
+        norm_const: constant used for normalization if save_format == 'normalized'
+    """
+
+    # get list of all fovs
+    fovs = list_folders(raw_data_dir, substrs=['fov'])
+
+    # load csv files
+    comp_mat = pd.read_csv(comp_mat_path, index_col=0)
+    panel_info = pd.read_csv(panel_info_path)
```
alex-l-kong, line 97: Same comment about file path validation here.

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+
+    if batch_size < 1 or not isinstance(batch_size, int):
+        raise ValueError('batch_size parameter must be a positive integer')
+
+    if gaus_rad < 0 or not isinstance(gaus_rad, int):
+        raise ValueError('gaus_rad parameter must be a non-negative integer')
+
+    # loop over each set of FOVs in the batch
+    for i in range(0, len(fovs), batch_size):
+        print("Processing image {}".format(i))
```
alex-l-kong, line 127: Would it be better to print the name of the FOV (ex. `print("Processing image {}".format(fovs[i]))` rather than the index?

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # blur data
+        if gaus_rad > 0:
+            batch_data_blurred = np.zeros_like(batch_data)
+            for j in range(batch_data.shape[0]):
+                for k in range(batch_data.shape[-1]):
+                    blurred = gaussian_filter(batch_data[j, :, :, k], sigma=gaus_rad)
+                    batch_data_blurred[j, :, :, k] = blurred
+        else:
+            batch_data_blurred = batch_data
```
alex-l-kong, line 142: If I'm not mistaken I don't think you need to store a separate array for `batch_data_blurred` and can just run `gaussian_filter` directly on `batch_data`:

```
if gaus_rad > 0:
    for j in range(batch_data.shape[0]):
        for k in range(batch_data.shape[-1]):
            batch-data[j, :, :, k] = gaussian_filter(batch_data[j, :, :, k], sigma=gaus_rad)
```

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+    # load images
+    dir_dict = {}
+    dir_shapes = []
+    for dir_name in input_dir_list:
+        dir_images = load_imgs_from_tree(dir_name)
+        dir_shapes.append(dir_images.shape)
+        dir_dict[dir_name] = dir_images
+
+    if not np.all([shape == dir_shapes[0] for shape in dir_shapes]):
+        raise ValueError("All directories must contain the same number of images")
```
alex-l-kong, line 241: I don't think we need `dir_dict` and `dir_shapes` for this and can simply validate using something along the lines of:

```
# get the number of images in the first FOV, would need to import list_files from io_utils
dir_len = len(list_files(input_dir_list[0], substr='.tiff'))

if not np.all([len(list_files(dir_len, substr='.tiff') == dir_len for dir_name in input_dir_list]):
    raise ValueError("All directories must contain the same number of images")
```

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # loop over each fov, and place into columns of tiled array
+        for i in range(first_dir.shape[0]):
+            start = i * img_size
+            end = (i + 1) * img_size
+
+            # go through each of the directories and place in appropriate spot
+            for idx, key in enumerate(input_dir_list):
+                tiled_image[(img_size * idx):(img_size * (idx + 1)), start:end] = \
+                    dir_dict[key].values[i, :, :, j]
```
alex-l-kong, line 261: Think we can remove `dir_dict` completely and simply do:

```
for idx, key in enumerate(input_dir_list):
    dir_images = load_imgs_from_tree(key)
    tiled_image[(img_size * idx):(img_size * (idx + 1)), start:end] = dir_images
```

alex-l-kong (changes requested):
Looks good, some comments related to file path validation and potential removal of some intermediate array creation procedures.

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+
+# TODO: Decide if this code is necessary or not
+# def compare_comped_images(raw_dir, comp_dir_list, output_dir):
+#     """Creates a tiled image containing the raw image, difference image, and output image
+#     from each compensated directory supplied
+#
+#     Args:
+#         raw_dir: directory containing raw images
+#         comp_dir_list: list of directories containing compensated images
+#         output_dir: directory where tifs will be saved"""
```
ackagel, line 184: similar to the tiling function below, some of this might be useful for other comparisons.  I'd say this can be a function, but maybe the bulk of it's image manipulation can be moved elsewhere (different folder, or part of the future "various utils from ark" package)

ackagel (commented):
looks good.  couple of suggested edits

creed/rosetta.py
```diff
@@ -0,0 +1,290 @@
...
+#             # subsequent rows are compensated images with different coefficients
+#             for idx, key in enumerate(comp_dir_list):
+#                 tiled_image[(img_size * (idx + 2)):(img_size * (idx + 3)), start:end] = \
+#                     comp_dict[key].values[i, :, :, j]
+#
+#         io.imsave(os.path.join(output_dir, raw_images.channels.values[j] + '_comparison.tiff'),
+#                   tiled_image)
+
+
+def create_tiled_comparison(input_dir_list, output_dir):
```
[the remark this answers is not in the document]
ngreenwald: Yeah, it's pretty flexible. Any sort of comparison across runs would benefit from this. Probably a good candidate for the toolbox repo once we create it. For now I'm going to keep it here since no one else is using it

creed/rosetta.py
ngreenwald: Yeah, this is basically a carbon copy of the above function, with the added step of creating a difference image between first and second supplied directory. I'm going to remove it, since I think the above function will get the job done, but if Cameron and I end up relying on this heavily then I'll add it back

creed/rosetta_test_cases.py
```diff
@@ -0,0 +1,48 @@
...
+class CompensateImageDataPanel:
+    # Create panel info pandas array
+
+    def case_sorted_panel(self):
+        # masses are sorted in numerical order, as expected
+        d = {'masses': test_mass_list, 'targets': test_chans_list}
+        panel_info = pd.DataFrame(d)
+        return panel_info
+
+    @xfail(raises=ValueError, strict=True)
```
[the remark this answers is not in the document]
ngreenwald: Yeah, found it reading the docs on pytest-cases

creed/rosetta.py
ngreenwald: These are the instrument names, meaning it's just fov1, fov2, etc

creed/rosetta.py
ngreenwald: This would require us to reload the entire directory for every iteration of the for loop. I agree it looks simpler, but the performance hit here would be significant, especially for larger directories

creed/rosetta.py
ngreenwald: If we didn't need `dir_dict` at all, I agree this would be a good fix. But if we're going to use it, I don't think this is any simpler than getting the shape as we loop through

creed/rosetta.py
alex-l-kong: Ah right, got a bit ahead of myself with this one.

alex-l-kong (approved):
Looks good to me!

ackagel (approved):
Looks good.  Will we add the default compensation matrix data itself in a different PR?

ngreenwald:
Yup, once we see how much modification is required

PR #30, opened 2022-01-26
Add TMA FOV validation functions for user sanity checking

### LINKED ISSUE #16: sanity checks for TMA remapping
In order to help prioritize user curation, we should create a couple distinct helper functions which check for likely errors in mapping. 

1. Distance between mapped centroids: This function will compute the distance between the centroids of each manual/automatic pair. It will then return the names of the 'x' FOVs with the greatest centroid distance, with the threshold set by the user.
2. Duplicate mapping: This function will return the names of any manual FOVs which map to the same automatic FOVs
3. Exact matching: This function will return the names of any manual FOVs whose name doesn't exactly match the name of the automatic FOV they are mapped to

### LINKED ISSUE #18: Ordering of automatically generated TMA FOV names
Currently, the names of the automatically generated FOV names are sorted in the same order that they are mapped. This makes it difficult to find a given name from a long list of names when trying to update the mapping. 

It would be better if the order of the names as displayed in the update widget was separate from the mapping order, so that they could be sorted alphabetically

alex-l-kong:

Closes #16 (and closes #18). Users may need some guidance as to which FOVs are the most likeliest that need remapping. This includes: 

1. User-proposed FOVs that are far away from their auto-generated FOV
2. User-proposed FOVs mapping to the same auto-generated FOV
3. User-proposed FOV names that don't match their corresponding auto-generated FOV names.

We want a way to display these to the user on the initial visualization, as well as update them on the fly when the user remaps FOVs.

We allow the user to specify which combination of the 3 issues stated above they wish to validate in `autolabel_tma_cores.ipynb`. Each of these issues is handled in a separate validation function that builds a list of mappings that do not pass the specified tests. These non-passes will be displayed in a text box right above the slide image display.


alex-l-kong:
Ok everything should be updated per our standup now. The validation text box is now just half of the original width.

I've also added tests for the dropdown menu functionality (finally).

alex-l-kong:
NOTE: now that we're running a lot of tests for certain functions randomizing order, need to start moving away from simple `!=` testing for randomization (the branch is failing again because of this, it'll happen more often than usual if we're running a double-digit amount of tests on these functions). I think the best way to handle this is to start using a random seed for these functions, although I'm open to other ideas.

alex-l-kong:
More along the lines of setting consistent random seeds, looks like the `pytest-randomly` library may allow us to do this fairly seamlessly: https://github.com/pytest-dev/pytest-randomly.

templates/autolabel_tma_cores.ipynb
```diff
@@ -195,6 +195,11 @@
     "* The selected proposed FOV and its mapped automatically-generated FOV are colored a darker shade.\n",
     "* The `Manually-defined FOV` menu can be used to visualize current mappings to automatically-generated FOVs.\n",
     "* The `Automatically-generated FOV` menu is used for re-mapping a FOV created by this script to one of yours.\n",
+    "* Additional validation parameters:\n",
```
ngreenwald, line 198: See if you can take another pass at these descriptions. Right now these feel like you're trying to write a manual. Reword them as if you're explaining to someone sitting next to you what each of these things does.

toffy/tiling_utils.py
```diff
@@ -928,12 +959,199 @@ def write_manual_to_auto_map(manual_to_auto_map, save_ann, mapping_path):
...
+    ) if check_duplicates else []
+    name_mismatch = find_manual_auto_name_mismatches(
+        manual_to_auto_map
+    ) if check_mismatches else []
+
+    # generate the annotation
+    warning_annot = ""
+
+    # add the manual-auto FOV pairs with distances greater than check_dist
+    # TODO: should we stick with pixels or should we convert to microns or stage coordinates?
```
ngreenwald, line 1107: I think microns makes the most sense as the distance metric

toffy/tiling_utils.py
```diff
@@ -947,8 +1165,18 @@ def interactive_remap(manual_to_auto_map, manual_fovs_info,
             the path to the file to save the mapping to
         draw_radius (int):
             the radius to draw each circle on the slide
+        check_dist (float):
```
ngreenwald, line 1168: duplicate

ngreenwald (changes requested):
Looks good, minor changes only

toffy/tiling_utils.py
alex-l-kong: Ok, I'll rework the distance computations in `assign_closest_fovs` to account for this.

alex-l-kong:
After some thought, I've decided to do some major restructuring to the pipeline:

- `assign_closest_fovs` will now be a hidden call in `tma_interactive_remap`...this cell was looking more and more unnecessary by the day and it'll help address the issue of the function being too involved.
- We'll have `assign_closest_fovs` return the original `manual_to_auto_map` and the `numpy.nparray` distance matrix (in microns) between each FOV. Having a lookup table for the distances between FOVs in microns will greatly reduce the number of conversion calls we'll need to make when ` remap_manual_to_auto_display` is called. It also removes the need for a `convert_pixels_to_microns` function when recalculating distances for validation purposes, preventing inconsistent results (due to floating point precision/rounding errors converting from microns to pixels and back won't yield the same results).

Let me know your thoughts @ngreenwald.

ngreenwald:
Sure thing, if you think it'll simplify things go for it

toffy/tiling_utils.py
```diff
@@ -590,54 +593,30 @@ def assign_closest_fovs(manual_fovs, auto_fovs):
...
 
     Returns:
         tuple:
 
-        - A `dict` defining the mapping of FOV names between `manual_fovs` and `auto_fovs`
-        - A `dict` defining each FOV in `manual_fovs` mapped to its centroid coordinates
-        - A `dict` defining each FOV in `auto_fovs` mapped to its centroid coordinates
+        - A `dict` mapping each manual FOV to an auto FOV and its respective distance from it
+        - A `pandas.DataFrame` defining the distance from each manual FOV to each auto FOV,
+          row indeices are manual FOVs, column names are auto FOVs
```
ngreenwald, line 603: typo

toffy/tiling_utils.py
```diff
@@ -928,19 +929,204 @@ def write_manual_to_auto_map(manual_to_auto_map, save_ann, mapping_path):
...
         manual_to_auto_map (dict):
             defines the mapping of manual to auto FOV names
-        manual_fovs_info (dict):
-            maps each manual FOV to its centroid coordinates
-        auto_fovs_info (dict):
-            maps each automatically-generated FOV to its centroid coordinates
+        manual_auto_dist (pandas.DataFrame):
+            defines the distance between each manual FOV from each auto FOV
+        dist_threshold (float):
+            if the distance (in microns) between a manual-auto FOV pair exceeds this value, it will
```
ngreenwald, line 942: specify units here of distance

ngreenwald (changes requested):
Nice, this is much more clear

templates/autolabel_tma_cores.ipynb
```diff
@@ -54,7 +54,7 @@
     "\n",
```
ngreenwald, line 54: <p>Four corners, not two</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/creed-helper/pull/30/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

toffy/tiling_utils.py
alex-l-kong: Not sure I follow, are you talking about something like `in stage coordinate microns`, or is this simply an example that should be used in other docstrings?

ngreenwald (approved):

PR #35, opened 2022-01-28
Normalize images by MPH

ngreenwald:

The goal of this PR is to implement the first part of the median pulse height (MPH) normalization code. For each image in a run, there will be slight variations in MPH, which decreases over time. We want to correct for this decrease in sensitivity by normalizing each image. Conceptually, there are two steps. 

First, we need to determine the relationship between MPH and signal. To do this, we construct a normalization curve that relates MPH to total counts. This entails extracting the median pulse heights, and total counts, from a range of different masses across images that were acquired with different MPH values. We can use this to form a titration curve to relate MPH to total counts. This is what the `combine_run_metrics` and `combine_tuning_curve_metrics` functions are for. The metrics that these functions are currently generated in a jupyter notebook: I'll move these into the codebase in the next PR. 

Once we have the total counts and the MPH of different masses, we can fit a curve. `fit_calibration_curve` and `create_objective_function` produce a fit for a given set of x and y values. This is used to fit the MPH to counts curve. 

However, not all masses in a given FOV will have the same MPH, due to differences in sensitivity. Therefore, for each FOV, we need to fit the relationship between mass and MPH. We use the extracted MPH counts and the masses to create a relationship between these two, and use it to generate predictions of the MPH for every mass in the dataset. These MPH values are then passed to the normalization function, which relates MPH to total signal, to produce a final value. This normalization value represents the fraction of maximum signal expected. We can then divide each image by this value to standardize it. 

**Remaining issues**

Future PRs will include: 

- The functions to compute the pulse height and total counts for each FOV (maybe integrated with the watcher script)
- A jupyter notebook with the overall workflow
- A settings.py style file for global variables, such as the masses extracted, across the different functions that are currently hardcoded


toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+    elif obj_func == 'poly_5':
+        def objective(x, a, b, c, d, e, f):
+            return a * x + b * x ** 2 + c * x ** 3 + d * x ** 4 + e * x ** 5 + f
+    elif obj_func == 'log':
+        def objective(x, a, b):
+            return a * np.log(x) + b
+    elif obj_func == 'exp':
+        def objective(x, a, b, c, d):
+            x_log = np.log(x)
+            return a * x_log + b * x_log ** 2 + c * x_log ** 3 + d
```
alex-l-kong, line 48: A dictionary with `lambda` functions might be more compact:

```
obj_functions = {
    'poly_2': lambda x, a, b, c: a * x + b * x ** 2 + c,
    ...
}
```

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+        plot_fit (bool): whether or not to plot the fit of the function vs the values
+
+    Returns:
+        list: the weights of the fitted function"""
+
+    # get objective function
+    objective = create_objective_function(obj_func)
+
+    # get fitted values
+    popt, pcov = curve_fit(objective, x_vals, y_vals)
```
alex-l-kong, line 69: With a `dict` of `lambda` objective functions, you can use `popt, _ = curve_fit(obj_functions[obj_func], x_vals, y_vals)`. Unless you plan on using `pcov` in a later PR I think it can safely be dropped.

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+    Returns:
+        func: prediction function"""
+
+    # get function based on specified pred type
+    obj_func = create_objective_function(name)
+
+    # define function which takes only x as input, uses weights for remaining parameters
+    def pred_func(x):
+        output = obj_func(x, *weights)
+        return output
```
alex-l-kong, line 97: With a `dict` of `lambda` objective functions, you could do this instead:

```
def pred_func(x):
    output = obj_functions[name](x, *weights)
    return output
```

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+
+    return pred_func
+
+
+def combine_run_metrics(run_dir, file_prefix):
```
alex-l-kong, line 102: Function needs a docstring.

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+                      'combined {} file in {}'.format(file_prefix, run_dir))
+        os.remove(os.path.join(run_dir, file_prefix + '_combined.csv'))
+        files = [file for file in files if 'combined' not in file]
+
+    if len(bins) != len(files):
+        raise ValueError('Mismatch between the number of bins and number '
+                         'of {} files'.format(file_prefix))
+
+    if len(bins) == 0:
+        raise ValueError('No bin files found in {}'.format(run_dir))
```
alex-l-kong, line 117: I think we can move this check right after defining the `bins` list.

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+        norm_func_path: file containing the saved weights for the normalization function
+    """
+
+    # get FOVs to loop over
+    if fovs is None:
+        fovs = io_utils.list_folders(data_dir)
+
+    # load calibration function
+    with open(norm_func_path, 'r') as cf:
+        norm_json = json.load(cf)
```
alex-l-kong, line 200: Might want to validate that `fovs` and `norm_func_path` are valid.

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # get images and pulse heights for current fov
+        images = load_utils.load_imgs_from_tree(data_dir, fovs=[fov], channels=channels,
+                                                dtype='float32')
+        fov_pulse_heights = pulse_heights.loc[pulse_heights['fov'] == fov, :]
+
+        # fit a function to model pulse height as a function of mass
+        mph_weights = fit_calibration_curve(x_vals=fov_pulse_heights['masses'].values,
+                                            y_vals=fov_pulse_heights['mphs'].values,
+                                            obj_func='poly_2')
```
alex-l-kong, line 221: Will the `obj_func` always be `'poly_2'`? If not, might want to specify that as an argument to this function.

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+def normalize_image_data(data_dir, output_dir, fovs, pulse_heights, panel_info,
+                         norm_func_path):
+    """Normalizes image data based on median pulse height from the run and a tuning curve
+
+    Args:
+        data_dir (str): directory with the image data
+        output_dir (str): directory where the normalized images will be saved
+        fovs (list or None): which fovs to include in normalization. If None, uses all fovs
+        pulse_heights: file containing pulse heights per mass per fov
+        panel_info_path: file containing mapping between channels and masses
```
alex-l-kong, line 190:
```suggestion
        panel_info (pandas.DataFrame): data containing mapping between channels and masses
```

alex-l-kong (changes requested):
Looks good, mostly minor suggestions. Also an alternative method for creating the objective functions in `create_objective_function`.

toffy/normalize_test.py
```diff
@@ -0,0 +1,196 @@
...
+
+@parametrize('plot_fit', [True, False])
+@parametrize('obj_func', ['poly_2', 'poly_3', 'poly_4', 'poly_5', 'log', 'exp'])
+def test_fit_calibration_curve(plot_fit, obj_func):
+    x_vals = np.random.rand(15)
+    y_vals = np.random.rand(15)
+    _ = normalize.fit_calibration_curve(x_vals, y_vals, obj_func, plot_fit)
+
+
+@parametrize('obj_func, num_params', [('poly_2', 3), ('poly_3', 4), ('poly_4', 5),
```
ackagel, line 44: this is one place where parametrizing a fixture, instead of the test case, can reduce the number of times these arguments are manually specified. not necessary to change imo, but something we should be aware of for future testing convention.

toffy/normalize_test.py
```diff
@@ -0,0 +1,196 @@
...
+                             'column_2': np.random.rand(10),
+                             'column_3': np.random.rand(10)}
+            metrics.append([metric_name, metric_values])
+
+        for bin_file in bins:
+            _make_blank_file(temp_dir, bin_file)
+
+        for metric in metrics:
+            name, values_df = metric[0], pd.DataFrame(metric[1])
+            values_df.to_csv(os.path.join(temp_dir, name), index=False)
```
ackagel, line 73: these can all be done in the same loop, no?

toffy/normalize_test.py
```diff
@@ -0,0 +1,196 @@
...
+        os.remove(os.path.join(temp_dir, bins[3]))
+        with pytest.raises(ValueError, match='Mismatch'):
+            normalize.combine_run_metrics(temp_dir, 'example_metric')
+
+        # empty directory raises error
+        empty_dir = os.path.join(temp_dir, 'empty')
+        os.makedirs(empty_dir)
+
+        with pytest.raises(ValueError, match='No bin files'):
+            normalize.combine_run_metrics(empty_dir, 'example_metric')
```
ackagel, line 106: I imagine these could be done via cases, but since we probably just need to crank this pr out now, this is fine.

toffy/normalize_test.py
```diff
@@ -0,0 +1,196 @@
...
+        dir_paths = [os.path.join(temp_dir, dir) for dir in dir_names]
+        combined = normalize.combine_tuning_curve_metrics(dir_paths)
+
+        # data may be in a different order due to matching dfs, but all values should be present
+        all_mph, all_counts = [], []
+        for df in mph_dfs:
+            all_mph.extend(df['mph'])
+
+        for df in count_dfs:
+            all_counts.extend(df['channel_counts'])
```
ackagel, line 130: the content in the loops can be merged together, no?

toffy/normalize_test.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # compute expected multiplier for FOV0
+        fov0_mults = mph_0 * weights[0] + weights[2]
+        fov0_mults = fov0_mults.reshape(1, 1, len(fov0_mults))
+        assert np.allclose(data_xr.values[0, :, :, :], normalized.values[0, :, :, :] * fov0_mults)
+
+        # compute expected multiplier for FOV1
+        fov1_mults = mph_1 * weights[0] + weights[2]
+        fov1_mults = fov1_mults.reshape(1, 1, len(fov1_mults))
+        assert np.allclose(data_xr.values[1, :, :, :], normalized.values[1, :, :, :] * fov1_mults)
```
ackagel, line 196: you could probably create everything via an `mph = (masses * 0.2 + 1, masses * .1 + 4)...` approach, and then loop over the assertions there.  Again, that's pretty much just polish/condensing and doesn't need-need to happen

ackagel (commented):
looks good, mostly just some polish comments

toffy/normalize.py
ngreenwald: Agreed. Remind me of this again on the next PR when we have the jupyter notebook included, then it will be obvious what should be validated where

toffy/normalize_test.py
ngreenwald: Yup, forgot I was going to create a `case` for this

toffy/normalize_test.py
ngreenwald: Is there an easy way to reuse the same data across different cases? When I tried to define variables within the `__init__` i got an error message. If so, I agree that this makes sense. However, it looks like I'd have to copy the code that makes the base case over in each of these, and then separately delete the data, which would clean up this test but just move the complexity to the cases file

toffy/normalize.py
```diff
@@ -0,0 +1,238 @@
...
+        fov_pulse_heights = pulse_heights.loc[pulse_heights['fov'] == fov, :]
+
+        # fit a function to model pulse height as a function of mass
+        mph_weights = fit_calibration_curve(x_vals=fov_pulse_heights['masses'].values,
+                                            y_vals=fov_pulse_heights['mphs'].values,
+                                            obj_func='poly_2')
+
+        # predict mph for each mass in the panel
+        mph_func = create_prediction_function(name='poly_2', weights=mph_weights)
+        mph_vals = mph_func(panel_info['masses'].values)
```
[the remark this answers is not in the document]
ackagel: Sure.  I guess I'm saying that `mph_func` can be retrieved **once**, outside the loop, directly from `create_objective_function` and be called via:

```
mph_vals = mph_func(panel_info['masses'].values, *mph_weights)
```

This creates only one, reused function object, while the current implementation creates a function object per fov.

toffy/normalize_test.py
ackagel: Yeah you can't specify anything in the `__init__` method (bummer).

The way I'd go about this is (kind-of) exemplified in `ark.utils.google_drive_utils_test`.  There's a `_fill_gdrive` method, as well as a `local_gdrive` decorator, which basically puts all tests into a pre-specified environment.  In the future 'various-utils' repo (which can include common test-utils), we can generalize this type of decorator to take a dictionary argument, which creates the specified environment (folders, files, etc.). 

For this purpose, a similar environment specifying decorator (or context class (using `with` keyword) ) can be used with the case class methods, as well as the `xfail` calls, and any parametrization.

Again, I'd be fine leaving it as is for now, but that's how I see file-i/o related testing being handled in the future

toffy/normalize.py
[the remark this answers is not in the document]
ngreenwald: Got it. I don't think there's any real efficiency gain, given how small these functions are. Do you think it's syntactically better to create a single function and update the weights within each iteration of the loop, rather than create a function with the weights within the loop? To me they look similarly interpretable.

toffy/normalize_test.py
ngreenwald: Sounds good. I converted to a case so that we can expand in the future.

toffy/normalize.py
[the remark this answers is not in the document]
ackagel: since this is qc, and run on a 'per-run' basis (~20 fovs), this doesn't matter much.  In terms of best practices/legibility, I think opting out of a function factory when possible is best, but yeah doesn't matter much either way.

alex-l-kong (commented):
Looks good to me, regarding testing, I just wanted to comment on something that Adam mentioned in a previous review comment regarding fixtures.

Working with Zak's interactive refactoring code, he has a pretty nice setup where he defines test fixtures in a `conftest.py` file inside the metaclustering folder. A library like `factory_boy` (https://github.com/FactoryBoy/factory_boy) can additionally be used for reusability among the test fixtures.

tl;dr we might want to go deeper down the testing iceberg.

toffy/normalize.py
```diff
@@ -0,0 +1,252 @@
...
+        combined['directory'] = dir
+        all_dirs.append(combined)
+
+    # combine data from each dir together
+    all_data = pd.concat(all_dirs)
+    all_data.reset_index()
+
+    # create normalized counts column
+    subset = all_data[['channel_counts', 'masses']]
+    all_data['norm_channel_counts'] = subset.groupby('masses').transform(lambda x: (x / x.max()))
```
ackagel, line 188: if we're normalizing by channel counts, an option to use the 99th percentile (instead of the max) to normalize could be useful.

ackagel (commented):
looks good.  just a q about normalization options

toffy/normalize.py
ngreenwald: Normally I would agree, but in this case since the max over the entire field, rather than an individual cell, it's fairly robust already.

PR #36, opened 2022-01-28
Ensure JSON data is read in with encoding UTF-8

### LINKED ISSUE #25: Moly QC point can't be written on windows computer

**Describe the bug**
The moly QC point includes a special character, the micron symbol, in one of the JSON fields. When this is read in, and then subsequently written to a JSON, it gets reformatted to include a special character, rather than the micron symbol, when used on the CAC. This causes the run JSON file to become improperly formatted. 

**To Reproduce**
Use moly point insertion functionality on CAC

image

alex-l-kong:

Closes #25. The Windows CAC uses a different default encoding when reading in JSON data. This present a problem for the Moly points in particular, which contain a unique µ symbol which gets garbled on Windows, causing the run files to become malformed. We need to override the default encoding on Windows so this character can be represented correctly.

Add an explicit `encoding='utf-8'` line to every `open` call that requires `json.load` or `json.dump`.

### DIFF
--- a/templates/autolabel_tma_cores.ipynb
+++ b/templates/autolabel_tma_cores.ipynb
@@ -112,7 +112,7 @@
     ")\n",
     "\n",
     "# save the automatically-named TMA FOVs to centroids mapping\n",
-    "with open(auto_fov_names_path, 'w') as afrp:\n",
+    "with open(auto_fov_names_path, 'w', encoding='utf-8') as afrp:\n",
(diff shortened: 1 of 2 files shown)


ngreenwald:
Looks good, I don't see the issue popping up. Will confirm later tonight when I load it into the mibi

ngreenwald (approved):

PR #41, opened 2022-02-07
Add tiled regions notebook to toffy

### LINKED ISSUE #40: Add tiled regions (non-TMA) notebook to toffy
**Section 1: Design details**

**Relevant background**

With the addition of the TMA tiling pipeline in `autolabel_tma_cores.ipynb`, we will need to add the tiled regions (non-TMA) notebook in. This notebook currently resides as `example_tiled_region_fov_generate.ipynb` in the `ark-analysis/templates_qc` folder.

**Design overview**

Since `tiling_utils.py` has been incorporated into `toffy`, all of the dependent functions are already there, so no major changes will need to be made to existing processes.

One issue that was recently addressed with the TMA tiling pipeline was the use of `x` and `y` suffixes, which is now deprecated in favor of `col` and `row` respectively. We'll need to make this change to the tiled region pipeline too.

Additionally, we'll remove the prompt for specifying the size of each FOV along the row and column axis, instead using the size values contained in the tiled region spec file provided by the user. Documentation will be included to ensure the user understands this is what's being done.

Finally, we should name the notebook `autolabel_tiled_region_cores.ipynb` to standardize naming conventions.

**Design list/flowchart**

No major changes to the pipeline.

**Required inputs**

Same inputs as needed for `example_tiled_regions_fov_generate.ipynb`.

**Output files**

Same outputs as generated by `example_tiled_regions_fov_generate.ipynb`.

**Control flow**

Only major change is removing the prompt for row and column FOV size for each region. The FOVs defining the top-left corner for each tiled region should contain a field called `fovSizeMicrons` which defines the dimensions along the row and column axis. This will be read in instead and used for the row and column FOV size.

At this point, the CAC does not support different dimensions along the row and column axes.

**Milestones and timeline**

No more than 2 days to integrate and test.

alex-l-kong:

Closes #40. Please consult the linked design doc for more info.

Consult #40.

### DIFF
--- a/templates/generate_tiled_regions.ipynb
+++ b/templates/generate_tiled_regions.ipynb
@@ -89,10 +89,10 @@
     "\n",
     "* The FOV defining the top-left corner and its centroid (taken from `fov_list_path`, each entry indicates a different region)\n",
     "* Number of FOVs along the x- and y-axis (user prompt)\n",
-    "* row- and column-axis step size (taken from the `fovSizeMicrons` field of each corner FOV in `fov_list_path`, same for row and col)\n",
+    "* row- and column-axis step size (taken from the `fovSizeMicrons` field of each corner FOV in `fov_list_path`, same for row and column)\n",
     "* Whether to randomize the order of the FOVs or not (user prompt)\n",
     "* Whether to insert Moly points between regions (user prompt)\n",
-    "* Whether to insert Moly points between FOVs at a specified interval (user prompt). Please refer to the documentation for `tiled_region_generate_fov_list` for how Moly point insertion works: https://ark-analysis.readthedocs.io/en/latest/_markdown/ark.mibi.html#ark.mibi.tiling_utils.tiled_region_generate_fov_list"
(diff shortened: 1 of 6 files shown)


templates/generate_tiled_regions.ipynb
```diff
@@ -0,0 +1,167 @@
+{
```
ngreenwald, line 1: <p>The moly path will always be the same, it's not run specific. Therefore, I think for this notebook, we don't need the set prefix. Instead, the output file name should default to be the input file name with <code>_tiled_fovs</code>  appended to the end</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/41/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ngreenwald (changes requested):
Looks good, just one update to the notebook

templates/generate_tiled_regions.ipynb
alex-l-kong: <p>Ok sounds good. I'll make the following changes:</p><p> </p><ul><li><code>region_corners_path</code>:  default to <code>region_corners.json</code>  without the prefix</li><li><code>tiled_region_fovs_path</code>: default to <code>region_corners_path + '_tiled_fovs.json</code> </li><li><code>moly_path</code>: default to <code>moly_point.json</code> <span class="ql-cursor">﻿</span></li></ul>

ngreenwald (approved):

PR #44, opened 2022-02-09
Add functionality to create a Rosetta matrix

alex-l-kong:

rosetta_matrices.py was added to the rosetta matrix branch. rosetta_matrices.py arguments: input csv file, output directory, list of multipliers, and channels to apply them. rosetta_matrices.py imports a csv file and applies to multipliers to the specified channels. If no channels are specified, all channels are multiplied by the multipliers. Output csv files are exported to the indicated directory. 

This helper function is designed for aiding titration to the compensation matrix for Betty. 

Created function: create_rosetta_matrix(default_matrix, output_directory, multipliers, channels = None). Function imports csv file and converts to pd dataframe. Checks channel argument - if none, it sets to all channels. If channels are not found in csv, raises value error. Iterates through multipliers, input matrix rows, and channels to multiply specified channels by multiplier. All non-affected channels are left unchanged. Output is exported to csv file in output directory. 

**Remaining issues**

Test function. Testing shows modifications aren't applied to the bottom row of input csv file. Need to address.


rosetta_matrices.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
+import pandas as pd
```
ngreenwald, line 2: You put this new version of the function in its own file. However, you should instead replace the previous version of your code that's in the `rosetta.py` file. Please move the code there, then delete this file

rosetta_matrices.py
```diff
@@ -0,0 +1,66 @@
...
+    """Creates a series of compensation matrices for evaluating coefficients
+    Args:
+        default_matrix (str): path to the rosetta matrix to use as the default
+        multipliers (list): the range of values to multiply the default matrix by
+            to get new coefficients
+        channels (list | None): an optional list of channels to include in the multiplication. If
+            only a subset of channels are specified, other channels will retain their values
+            in all iterations. If None, all channels are included
+
+    >>> create_rosetta_matrices('compensation_matrix_intensity_nosum.csv', r'/Users/cameron/PycharmProjects/RosettaBetty', [0.5, 1, 1.5], [117])
```
ngreenwald, line 14: Remove the docstest, we'll switch to doing full function testing

rosetta_matrices.py
```diff
@@ -0,0 +1,66 @@
...
+    for i in channels:
+        if i not in comp_channels:
+            raise ValueError('Specified channel does not exist')
+
+    # step 3: loop over each of the multipliers (anything the user inputs, make an output matrix)
+    # step 4: modify the appropriate channels based on mulitiplier
+    # step 5: save the modified matrix as original_name_multiplier
+    for i in multipliers: # returns each comp_matrix value
+        zero_matrix = np.zeros(shape=(matrix_rows+1, matrix_columns))
+        modified_matrix = pd.DataFrame(zero_matrix[1:], index = row_labels, columns=column_features)
```
ngreenwald, line 39: This was my fault, instructions weren't clear. For channels that aren't specified by the user, we want the row corresponding to that channel to be blank, not the column.

For example, if 168 isn't included, the entire row corresponding to 168 should be zeros for all of the columns

rosetta_matrices.py
```diff
@@ -0,0 +1,66 @@
...
+            for k in channels:
+                if k in comp_channels:
+                    channel_index = comp_channels.index(k)
+                    modified_matrix.iloc[channel_index,0:] = comp_matrix.iloc[channel_index,0:] * i
+            modified_matrix.iloc[j,0:] = comp_matrix.iloc[j,0:]
+        df = pd.DataFrame(modified_matrix)
+        df.to_csv(output_directory+'/Rosetta_Titration%s.csv' % (str(i)))
+
+
+def test_create_rosetta_matrices():
```
ngreenwald, line 50: Move this test function to the dedicated testing function that alex started for you in rosetta_test.py

rosetta_matrices.py
```diff
@@ -0,0 +1,66 @@
...
+    df2 = pd.DataFrame(random_matrix)
+    df2.to_csv(r'/Users/cameron/PycharmProjects/RosettaBetty/Random_matrix.csv')
+    #test_matrix = pd.DataFrame(random_matrix[0:])
+
+
+    # step 3: run create_rosetta_matrices using template matrix
+    create_rosetta_matrices('Random_matrix.csv', r'/Users/cameron/PycharmProjects/RosettaBetty', [2])
+
+    # step 4: check that output is correct
+    test = pd.read_csv('Rosetta_Titration2.csv', index_col=0)  # pandas DataFrame
```
ngreenwald, line 64: To "check" that the output is correct, we need to use an `assert` statement. 

See here for an example: https://github.com/angelolab/toffy/blob/4b684e8b2759229d581b2c0914f97e7044e719e2/toffy/file_hash_test.py#L12

ngreenwald (changes requested):
See below for my comments. I find it's easiest to view these on github directly

cwalker8:
Last row of comp_matrix is still not affected by multipliers (output of last row is all zeroes). All other comments addressed

toffy/rosetta_test.py
```diff
@@ -217,3 +217,27 @@ def test_replace_with_intensity_image(overwrite, folders):
...
+
+
+    # step 3: run create_rosetta_matrices using template matrix
+    create_rosetta_matrices('Random_matrix.csv', r'/Users/cameron/PycharmProjects/RosettaBetty', [2])
+
+    # step 4: check that output is correct
+    test = pd.read_csv('Rosetta_Titration2.csv', index_col=0)  # pandas DataFrame
+    out = test/random_matrix
+    matrix_rows = len(out)
+    matrix_columns = len(out.iloc[0])
```
ngreenwald, line 239: check out `np.array_equal` instead of comparing element by element

toffy/rosetta.py
```diff
@@ -302,25 +302,45 @@ def replace_with_intensity_image(base_dir, channel, replace=True, folders=None):
...
+
+    # step 3: loop over each of the multipliers (anything the user inputs, make an output matrix)
     # step 4: modify the appropriate channels based on mulitiplier
-
     # step 5: save the modified matrix as original_name_multiplier
+    for i in multipliers: # returns each comp_matrix value
+        zero_matrix = np.zeros(shape=(matrix_rows+1, matrix_columns))
+        modified_matrix = pd.DataFrame(zero_matrix[1:], index = row_labels, columns=column_features)
+        for j in range(matrix_rows):
+            for k in channels:
```
ngreenwald, line 337: The logic here is a bit muddled. 

You first loop over each `j` row, where each row corresponds to a channel. 
However, after you've picked the `jth` row, you then loop over each channel, and for every channel, check if it's part of comp_channels. 

You then repeat this same thing for every time you pick a new row. At the end of this, you then multiply the whole thing by zero?

ngreenwald (changes requested):

ngreenwald (changes requested):
Tests are still failing, looks like you forgot to remove the multiplier

toffy/rosetta.py
```diff
@@ -300,27 +300,45 @@ def replace_with_intensity_image(base_dir, channel, replace=True, folders=None):
...
+    matrix_columns = len(comp_matrix.iloc[0])
+    column_features = list(comp_matrix.columns.values)
 
-    # step 3: loop over each of the multipliers
+    # step 2: figure out which channels will be modified (rows)
+    if channels is None:
+        channels = comp_channels
 
-    # step 4: modify the appropriate channels based on mulitiplier
+    for i in channels:
```
ngreenwald, line 328: This check only needs to happen if the user has provided channels; otherwise, channels is already equal to comp_channels. This can be put as an `else` statement that gets paired with `if channels is None`

toffy/rosetta.py
```diff
@@ -300,27 +300,45 @@ def replace_with_intensity_image(base_dir, channel, replace=True, folders=None):
...
+        if i not in comp_channels:
+            raise ValueError('Specified channel does not exist')
 
+    # step 3: loop over each of the multipliers (anything the user inputs, make an output matrix)
+    # step 4: modify the appropriate channels based on mulitiplier
     # step 5: save the modified matrix as original_name_multiplier
-
-    pass
+    for i in multipliers:  # returns each comp_matrix value
+        zero_matrix = np.zeros(shape=(matrix_rows + 1, matrix_columns))
```
ngreenwald, line 336: You make the zero matrix have shape rows + 1, but then in the following line you skip the first row. I think we can just make it the correct size to start with, right?

toffy/rosetta.py
```diff
@@ -300,27 +300,45 @@ def replace_with_intensity_image(base_dir, channel, replace=True, folders=None):
...
+    # step 2: figure out which channels will be modified (rows)
+    if channels is None:
+        channels = comp_channels
 
-    # step 4: modify the appropriate channels based on mulitiplier
+    for i in channels:
+        if i not in comp_channels:
+            raise ValueError('Specified channel does not exist')
 
+    # step 3: loop over each of the multipliers (anything the user inputs, make an output matrix)
```
ngreenwald, line 332: Remove the comments I left for you as a guidepost, and instead comment the code the way you did the test functions to explain what's happening

toffy/rosetta_test.py
```diff
@@ -217,3 +218,58 @@ def test_replace_with_intensity_image(overwrite, folders):
...
+        multipliers = [4]
+        channels = [197]
+        create_rosetta_matrices(output_path, temp_dir, multipliers, channels)
+
+        rosetta_path_2 = os.path.join(temp_dir, 'Rosetta_Titration%s.csv'
+                                      % (str(multipliers[0])))
+        # grabs output of create_rosetta_matrices
+        test_matrix_2 = pd.read_csv(rosetta_path_2, index_col=0).astype(int)
+        # row index of input channel
+        test_index = test_channels.index(channels[0])
```
ngreenwald, line 261: Instead of checking that the specified row has been modified by the multiplier and looping over all of the other rows, you could divide the `test_matrix` by a vector that has 1 for every channel which isn't supposed to be modified, and the multiplier for the channels that are supposed to be modified. Then you could use the same check as above, making sure the validation is equal to the random_matrix.

toffy/rosetta_test.py
```diff
@@ -217,3 +218,58 @@ def test_replace_with_intensity_image(overwrite, folders):
...
+        # grabs output of create_rosetta_matrices
+        test_matrix = pd.read_csv(rosetta_path, index_col=0).astype(int)
+        validation = (test_matrix / multipliers[0])
+
+        # confirm all channels scaled by multiplier
+        assert np.array_equal(random_matrix, validation)
+
+        # Checking Specific Channels
+        multipliers = [4]
+        channels = [197]
```
ngreenwald, line 253: I would change this to test two channels, rather than just one, to be more comprehensive.

ngreenwald (changes requested):
Nice job! All the tests are passing. See below for my comments on some further cleanup.

ngreenwald (approved):
Awesome, looks great!

PR #45, opened 2022-02-16
Add notebook to change co-registration parameters in settings.py

alex-l-kong:

Currently, we have fixed constants for co-registering coordinates from `fovCenterMicrons` in each FOV of a spec file to the slide for visualization. However, for the stage coordinates to optical pixels conversion, this will fail if the slide position gets moved on the commercial instrument, causing the fiducial positions to change.

We add a notebook that allows the users to re-align the co-registration process with the new fiducial locations.

The notebook, `update_coregistration_parameters.ipynb`, allows the user to specify the top 3 fiducials on the slide in both stage coordinates and optical pixel coordinates. These can be retrieved from the CAC.

Because the x- and y-coordinate axes are on different scales, we run separate regressions from stage coordinate to optical pixel coordinate values to generate the multiplier and offset values for both.

The multiplier is simply the slope of the regression. The offset is the x-intercept * -1: this is mostly a stylistic choice and `y = m(x + (x-intercept * -1))` is the same as `y = mx + b`. I feel it's a bit easier to understand in this context that we need to offset off of the stage coordinate then add a multiplication factor to retrieve the corresponding optical pixel coordinate. The concept of a y-intercept (b) is a little less intuitive in this context since no one will ever have a FOV with a stage coordinate of 0.

**Remaining issues**

Some possible aesthetic changes:

- Having a user prompt for the stage and pixel coordinate values of the fiducials
- Moving some processes to functions in `tiling_utils.py`

### DIFF
--- a/templates/update_coregistration_params.ipynb
+++ b/templates/update_coregistration_params.ipynb
@@ -1,0 +1,204 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ec35cb18-c771-48e6-859f-e10e074545d6",
+   "metadata": {},
+   "source": [
+    "# Update coregistration constants"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "edb13b4a-9589-4c2d-814c-0cae835cc255",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


ngreenwald:
Also update the logic requiring R1C1 naming

ngreenwald (changes requested):
Nice, looks good. Can you update the autolabel_tma notebook with a note saying that co-registration needs to be performed the first time the notebook is run on a new mibi?

PR #50, opened 2022-03-01
Add watcher compatible QC notebook

### LINKED ISSUE #46: Refactor QC metrics to work with watcher function
**Section 1: Design details**

The current QC metric notebook/workflow assumes that the user will run it once their run is complete. Moving forward, we want this to be run automatically as part of the watcher script.

We should refactor QC metrics to work with a watcher function which operates on a single FOV at a time, saving the output.

**Design overview**

The QC metric pipeline consists of 3 parts that will need varying degrees of modification:

1. Download raw images from MIBItracker: this will now be handled by `mibi_bin_tools`. Users will need to specify a path to their `.bin` files and this step will be done on a FOV-by-FOV basis.
2. QC metric extraction: this will also be done on a FOV-by-FOV basis and will append directly to files rather than to in-memory DataFrames. **We will also need to store the run order in a separate file.**
3. QC metric visualization: for now, we'll leave this to run after all the FOVs have been downloaded and QC-extracted. In the future, this may transition to a real-time FOV-by-FOV basis.

**Design list/flowchart**

watcher_qc drawio

**Required inputs**

- The path to the bin file folder (contains the MIBI `.bin` files). Each folder will contain a set of files prefixed with the FOV in question. For example, `fov-1-scan-1`.
- **While a bin folder may contain multiple FOVs, only 1 FOV will ever be computed at a time in `compute_qc_metrics`**. The specific FOV that will be computed needs to be passed as an argument to `compute_qc_metrics` from either the watcher script or `example_qc_metrics_comp.ipynb`. **The name of the FOV is defined by the corresponding JSON file (for example, if `fov-1-scan-1.json` is in the folder, the name of the FOV needs to be specified as `fov-1-scan-1`) and not just `fov-1`**.
- The path to the panel file. This defines the integration ranges the bin file needs to extract properly.
- The path to the QC metric folder (contains the `.csv` files for each QC metric per FOV, as well as the FOV run order)
- The path to the visualization folder (contains the `.png` files visualizing each QC metric sorted by run order of FOVs)

**Output files**

The data that gets placed in the QC metric and visualization folders described above.

**Control flow**

For each QC pipeline step outlined in the `Design overview` portion of Section 1, the corresponding functions will need varying degrees of modification:

1. `compute_qc_metrics` will now compute on just a single FOV per iteration. It will take in an argument `bin_fie_name` which will specify a single FOVs data. Only this FOV will be passed to `mibi_bin_tools.extract_bin_files` with the user-provided bin file folder as `data_dir` without specifying an `out_dir`. The result will be an `xarray.DataArray` containing the channel images for the FOV.
- When calling `extract_bin_files`, the resulting `DataArray` will contain a dimension called `'type'`: for the purposes of this function, this will only contain one coord (`'pulse'`) and is what we'll be subsetting on.

2. To ensure that the watcher script and `example_qc_metrics_comp.ipynb` know which FOVs have been run or not, `compute_qc_metrics` at the end saves a unique file for the FOV passed in with the corresponding QC metric values for each FOV. For non-zero mean intensity, an example file for `fov0` might be called `fov0_nonzero_mean_stats.csv` and look like:

| fov | channel | metric_name |
| ---- | ------ | ----- |
| 'fov0' | chan0 | val1 |
| 'fov0' | chan1 | val2 |
| ... | ... | ... |

We save the data in melted format to make aggregation easier and for visualization purposes. This data will be saved in a specific folder; the path will need to be provided to the function as `qc_dir`.

3. We add an aggregation function, `combine_qc_metrics`, which will simply append all the files for each QC metric together and create one dataset. We can return this as a `pd.DataFrame` or save as a `.csv`, whichever is easier.

4. The watcher script and `example_qc_metrics_comp.ipynb` will worry about which FOVs to pass in. For the latter, the user will need to specify the path to their bin file (called `bin_file_path`) as well as the `qc_dir` they wish to write to. A loop inside the notebook will then check each FOV in `bin_file_path` and see if the corresponding FOV QC metric files exists in `qc_dir`. If they do not, it will pass in that FOV to `compute_qc_metrics` which then runs the QC extraction for that FOV and saves it.
 
5. `visualize_qc_metrics` can stay as is: a `save_dir` can be provided which will store the images created for each metric.

**Milestones and timeline**

1. `mibi_bin_tools` integration: shouldn't take more than 1 day
2. `qc_metric` refactoring: 1 day
3. Testing the updated pipeline: 1-2 days

alex-l-kong:

Closes #46. Use the issue link to see more details.

See #46.

**Remaining issues**

Possibly need to refactor based on how the watcher script will feed the inputs in.


alex-l-kong:
Most bin files will contain just a single FOV, however the new `compute_qc_metrics` also supports bin files with multiple FOVs (one of the tests in `mibi_bin_tools` has 2 FOVs).

alex-l-kong:
I've decided not to move the QC visualization from the notebook for now.

ngreenwald (changes requested):
Alex, I don't understand what happened with this PR. The point was to make this a modular workflow that computes the QC metrics for one FOV at a time. This is spelled out explicitly in the design doc you made. 

However, the `compute_qc_metrics` function you wrote loops over all FOVs in the directory, rather than working on one FOV at a time. It doesn't name each qc metric file according the fov it's derived from. It doesn't have an aggregation function in the notebook to integrate the metrics from different FOVs. 

All of this is spelled out accurately in the design doc, but didn't make it in the PR.

alex-l-kong:
@ngreenwald let's meet sometime tomorrow to discuss. Here's the rundown from my end:

1. My OG implementation was single-FOV only; I would've left it this way if only the watcher script called `compute_qc_metrics`. The motivation for re-adding the FOV loop back was due to the nature of one of the `mibi_bin_tools` test cases (copied to `toffy` under `data`). The `moly` test folder contains multiple FOVs (`fov-1-scan-1` and `fov-2-scan-1`). I'm concerned over the case where a user needs to run `example_qc_metric_comp.ipynb` directly and specified a multi-FOV bin file expecting all FOVs to be QC-extracted. I had a bad feeling in my leg that this could be problematic if we didn't address this now, thus the last-minute change.

2. I think the bullet point `The path to the QC metric folder (contains the `.csv` files for each QC metric per FOV per channel...` potentially needs revisiting. I was meaning to refer to the aggregated `.csv` file created for each QC metric defining the computed values for each FOV (and not to separately-named FOV files per QC metric). In other words, 3 `.csv` files in the QC metric folder–`nonzero_mean_stats.csv`, `total_intensity_stats.csv`, and `percentile_99_9_stats.csv`–that look like (both original and updated implementation):

| chan1 | chan2 | fov |
| ------ | ------ | --- |
|  qc_metric_val_fov_1_chan_1 | qc_metric_val_fov_1_chan_2 | fov1 |
|    ...     |    ...     | ...  |

When a new FOV comes in, we compute the QC metrics on the corresponding image and per the design doc, append the computed values directly to the files (`nonzero_mean_stats.csv`, `total_intensity_stats.csv`, `percentile_99_9_stats.csv`).

Let me know if anything else in the design doc (and thread) wasn't clear in this regard. I'm also open to continued discussion over alternative storage options.

3. Did point 2 address the issue of aggregation, or were you thinking of something else?

ngreenwald:
Yeah probably face-to-face will be easiest, lets discuss after standup

alex-l-kong:
Updated `qc_comp` to the new design doc.

alex-l-kong:
In summary:

- `compute_qc_metrics`: extracts a single FOV in `bin_file_path`, writes individual fov-level metrics (`{fov_name}_nonzero_mean_stats.csv`, `{fov_name}_total_intensity_stats.csv`, and `{fov_name}_percentile_99_9_stats.csv`) to `bin_file_path`
- `combine_qc_metrics`: for each QC metric, takes all the fov-level files and combines them in `combined_{metric_name}_stats.csv`. The FOVs are ordered in run-acqisition order (determined by the number associated with it)
- `example_qc_metric_comp.ipynb`:
    - Reads all the FOVs in from `bin_file_path`, however instead of passing all into `compute_qc_metrics`, first loops over each, checks if the fov-level QC metric files exist in `bin_file_path`, and if not, then runs it through `compute_qc_metrics`.
    - Adds an argument `generate_qc_graphs` that if set, will run the aggregation function (`combine_qc_metrics`), then run the resulting visualizations

toffy/qc_comp.py
```diff
@@ -246,209 +248,163 @@ def compute_99_9_intensity(image_data):
...
 
+    # save the data to the qc_dir
+    df_nonzero_mean.to_csv(
+        os.path.join(bin_file_path, '%s_nonzero_mean_stats.csv') % fov_name, index=False
+    )
+    df_total_intensity.to_csv(
+        os.path.join(bin_file_path, '%s_total_intensity_stats.csv') % fov_name, index=False
+    )
+    df_99_9_intensity.to_csv(
+        os.path.join(bin_file_path, '%s_percentile_99_9_stats.csv') % fov_name, index=False
```
ngreenwald, line 355: Can the df manipulation be condensed into a for loop? We're doing the same operation on all three dfs

toffy/qc_comp.py
```diff
@@ -246,209 +248,163 @@ def compute_99_9_intensity(image_data):
...
-        '99_9_intensity': df_99_9_intensity
-    }
-
-    return qc_data
+    # path validation check
+    if not os.path.exists(bin_file_path):
+        raise FileNotFoundError('bin_file_path %s does not exist' % bin_file_path)
+
+    # iterate over each metric substr
+    metric_substrs = [
```
ngreenwald, line 372: Yeah, can we do something like this in the function above?

toffy/qc_comp.py
```diff
@@ -246,209 +248,163 @@ def compute_99_9_intensity(image_data):
...
+        # define an aggregated metric DataFrame
+        metric_df = pd.DataFrame()
+
+        # list all the files corresponding to this metric
+        metric_files = io_utils.list_files(bin_file_path, substrs=ms)
+
+        # this prevents an already-existing combined .csv file from sneaking in
+        metric_files = [
+            mf_match[0] + '_%s' % ms for mf in metric_files if
+            len(mf_match := re.findall(r'fov-\d+-scan-\d+', mf)) == 1
```
ngreenwald, line 385: Can we remove files that have 'combined' in them instead?

toffy/qc_comp.py
```diff
@@ -246,209 +248,163 @@ def compute_99_9_intensity(image_data):
...
+        ]
+
+        # sort the files to ensure consistency
+        # NOTE: all FOVs are named fov-m-scan-n, 'm' determines run acquisition order
+        # TODO: need an additional sort by scan?
+        metric_files = sorted(metric_files, key=lambda x: re.findall(r'\d+', x)[0])
```
ngreenwald, line 391: We should pull this out into a helper function that works from `m` from 1 to 101 and `n` from 1 to 99. Other functions will need to sort FOVs by order

toffy/qc_comp_test.py
```diff
@@ -178,129 +181,170 @@ def test_compute_99_9_intensity():
...
+
+        # read in the QC data
+        nm = pd.read_csv(nonzero_mean_path)
+        ti = pd.read_csv(total_intensity_path)
+        p99_9 = pd.read_csv(percentile_99_9_path)
+
+        # assert the column names are correct
+        assert list(nm.columns.values) == ['fov', 'channel', 'Non-zero mean intensity']
+        assert list(ti.columns.values) == ['fov', 'channel', 'Total intensity']
+        assert list(p99_9.columns.values) == ['fov', 'channel', '99.9 intensity value']
```
ngreenwald, line 255: Can this stuff be condensed into a for loop over the tuple of file names and column names?

toffy/qc_comp_test.py
```diff
@@ -178,129 +181,170 @@ def test_compute_99_9_intensity():
...
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'random.csv'))
+        Path(os.path.join(bin_file_path, 'random.txt')).touch()
+
+        # add existing combined .csv files, this also should not affect aggregation
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'combined_nonzero_mean_stats.csv'))
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'combined_total_intensity_stats.csv'))
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'combined_percentile_99_9_stats.csv'))
+
+        # define sample dataframes for each fov to write to qc_path
+        for fov in fovs:
```
ngreenwald, line 286: More condensing of repeated stuff per metric file

toffy/qc_comp_test.py
```diff
@@ -178,129 +181,170 @@ def test_compute_99_9_intensity():
...
+        bin_file_path = os.path.join(temp_dir, 'sample_qc_dir')
+        os.mkdir(bin_file_path)
+
+        # put some random stuff in bin_file_path, test that this does not affect aggregation
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'random.csv'))
+        Path(os.path.join(bin_file_path, 'random.txt')).touch()
+
+        # add existing combined .csv files, this also should not affect aggregation
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'combined_nonzero_mean_stats.csv'))
+        pd.DataFrame().to_csv(os.path.join(bin_file_path, 'combined_total_intensity_stats.csv'))
```
ngreenwald, line 282: Why do we need this at the beginning?

toffy/qc_comp_test.py
```diff
@@ -178,129 +181,170 @@ def test_compute_99_9_intensity():
...
+        assert os.path.exists(combined_ti_path)
+        assert os.path.exists(combined_p99_9_path)
+
+        # read in the QC data
+        nm = pd.read_csv(combined_nm_path)
+        ti = pd.read_csv(combined_ti_path)
+        p99_9 = pd.read_csv(combined_p99_9_path)
+
+        # assert the column names are correct
+        assert list(nm.columns.values) == ['fov', 'channel', 'Non-zero mean intensity']
```
ngreenwald, line 337: If you create the fake qc data programmatically, rather than randomly, we can assert that it the values are correct as well.

ngreenwald (changes requested):
Looks good, there's a lot of areas where the repetition between the three different metrics can be condensed but logic is all there

toffy/qc_comp.py
alex-l-kong: In other words, toss out any bin file FOV name that has a fov number outside `range(1, 102) and scan number outside `range(1, 100)` when sorting?

toffy/qc_comp_test.py
alex-l-kong: We write these blank `combined.csv` files to ensure that even if `mibi_bin_path` contains them, `combine_qc_metrics` won't try to grab these files when aggregating. Specifically, it tests this line: `metric_files = [mf for mf in metric_files if 'combined' not in mf]`.

toffy/qc_comp.py
ngreenwald: It shouldn't out numbers outside that range, but it should definitely work for numbers inside that ranges. I don't expect we'll ever have four digits worth of bin files or three digits worth of scans

toffy/qc_comp_test.py
alex-l-kong: I've done this for `combine_qc_metrics`. However it will be difficult to do with `compute_qc_metrics` as none of the test `.bin` files especially with the Gaussian blurring tests.

templates/example_qc_metric_eval.ipynb
```diff
@@ -3,6 +3,7 @@
   {
```
ngreenwald, line 3: <pre class="ql-syntax" spellcheck="false">Line #3.    fovs = [fov_match[0] for f in fovs if len(fov_match := re.findall(r'fov-\d+-scan-\d+', f)) == 1]
</pre><p>Why do these need to be sorted? I thought the aggregation function handles that</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/50/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ngreenwald, line 3: <p>What is the point of this flag? The cells below don't contain any other code, so either the user runs them and generates the graphs, or doesn't run them and doesn't generate the graphs</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/50/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

ngreenwald (changes requested):
Looks good, just a little cleanup on the notebook

templates/example_qc_metric_eval.ipynb
alex-l-kong: <p>This isn't a sorting function, it was to prevent extra <code>.json</code> files (ex. <code>2022-01-22_adam_control_tissue_2675V.json</code> in <code>2022-01-22_adam_control_tissue_2675V</code>) from making it in. That being said, I can just use the <code>.bin</code> files to extract the names. </p>

alex-l-kong: <p>Just an additional control flow flag, I'll remove it.</p>

alex-l-kong: Oh I see now, I'll also need to adjust the comment.

ngreenwald (approved):
Looks good!

Issue #28, opened 2022-01-24
Watcher script


ackagel:
### **Relevant background**

> Give a brief (1-2 paragraph) description of the problem/issue that this feature is solving, and how it addresses the issue

Some scripts may take awhile to run per FOV.  It would save time if these scripts could be scheduled to run as soon as a FOV was ready for processing.

### **Design overview**

> Describe each of the main components of the proposed project

Primarily a script which can observe a folder.  When a new FOV is added to the folder (json + bin) the script should immediately begin calling user-specified processing functions on that FOV.

### **Design list/flowchart**

>Provide an overview of how the different pieces of the project will interact with each other, either in a numbered list, bullet point, or brief sketch

 1. User specifies folder to watch (should only need run name)
 2. User specifies functions to call/chain on each FOV-completion event
 3. When run, the script parses folder for run json (gets number of fovs, other metadata)
 4. On complete FOV (json + bin), the script runs the specified functions
 5. Watcher function exits once all expected FOVs are complete.

### **Required inputs**

> Provided a description of the required inputs for this project, including column names for dfs, dimensions for image data, prompts for user input, etc

Just a run name (technically a folderpath) for the base script.  For the chained function calls, other inputs might be required depending on the function...

### **Output files**

>Provide a description of the outputs for this project, including column names for dfs, dimensions for image data, etc. If any plots will be generated, provide (simple) sketches demonstrating the plot type and axes labels. 

At a minimum a log file.  Full outputs depend on the specified function callbacks, but generally extracted TIFs, QC metrics, etc.

## **Implementation details**

> Once you have completed section 1, please tag the relevant parties and iterate on the initial design details until everyone is satisfied. Then, proceed to section 2

### **Control flow**

> Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

### **Milestones and timeline**

> List each of the major components of the project, and provide an estimated completion date for each one.

ngreenwald:
Great, looks good

ngreenwald:
This is what happens when the files are in the process of transferring 
![image](https://user-images.githubusercontent.com/13770365/150881178-868dfa18-7e44-4c23-88e9-605dcfc7ad97.png)

ackagel:
remind me, is the run json transferred over before data aq?

Also, is it possible to generate a checksum/hash of the completed bin/json on the instrument computer before the transfer?  That could be a good way of removing any risk of early file reading/processing.

ngreenwald:
[...]

PR #53, opened 2022-03-03
adds fov_watcher template

ackagel:
## **Purpose**

Adds watcher as described in issue #28 

## **Implementation**

Adds `fov_watcher.py`.  Watcher is built on top of the watchdog package.

## **Remaining issues**

Notebook integration + tests.  Also, the callback function signatures might need to be changed to bin file paths, instead of the current signature requiring the xarray format.  The former is preferable, as it allows callback functions to use individual/internal panels.


toffy/fov_watcher.py
```diff
@@ -0,0 +1,217 @@
...
+    
+        Returns:
+            (bool, str):
+                whether or not both json and bin files exist, as well as the name of the point
+        """
+
+        # TODO: check watchdog path depth
+
+        fov_name, extension = path.split('.')[0:2]
+        if fov_name in self.completion:
```
ngreenwald, line 61: If `fov_name` isn't in the `completion` dict we should raise some type of error, since it's initialized with everything from the directory right?

toffy/fov_watcher.py
```diff
@@ -0,0 +1,217 @@
...
+        This function is automatically called; users generally shouldn't call this function
+
+        Args:
+            event (FileCreatedEvent):
+                file creation event
+        """
+        super().on_created(event)
+
+        # check if what's created is in the run structure
+        fov_ready, point_name = self.run_structure.check_run_condition(event.src_path)
```
ngreenwald, line 143: I think this will need to be modified to account for the fact that files are created with size zero, and then later copied over.

toffy/fov_watcher.py
```diff
@@ -0,0 +1,217 @@
...
+            log_file_path = os.path.join(self.watcher_out, 'log.txt')
+            logf = open(log_file_path, 'a')
+
+            logf.write(
+                f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- '
+                f'Extracting {point_name}'
+            )
+
+            # extract data
+            img_data = \
```
ngreenwald, line 154: If we have multiple functions that require extracted data then this makes sense, otherwise it may be easier to have each watcher function handle its own processing

ngreenwald (changes requested):
Looks good so far, see below for comments

toffy/fov_watcher.py
ackagel: right now, it's only checking for bin/json files. I'd say it should only raise an error if `fov_name` isn't a key in `completion`, **and** `extension` is `'bin'`.

toffy/fov_watcher.py
ackagel: for sure; I think I can just set a small delay (like 10s?) to wait for file download completion. since this already runs asynchronously, a delay shouldn't interfere with the file watching or further queued callbacks.

toffy/fov_watcher.py
ackagel: honestly, even just for the 'bonus' of having each watcher function have its own extraction parameters would probably be handy (intensities y/n, integration range checks, etc).  I'll change it to pass the run folder + fov name instead.

toffy/fov_watcher.py
ngreenwald: Sometimes it takes seconds, sometimes it's like 3 minutes. There's never an in between; it's either size zero or completely copied over. Can you check that the file has a non-zeo size?

ackagel: oh awesome, yeah that's easy. I might add a timeout argument (default like 10mins or something), just so things don't hang if something never arrives

ngreenwald: Yup good idea, sometime the first one will never copy over

ackagel:
looks like tests are failing because of a breaking change from an xarray minor version update... submitted a fix to `mibi-bin-tools`. should be ready for review regardless; only thing missing is testing for the file watching functionality itself (as opposed to the callbacks).

srivarra:
Do you think it would be valuable to put a some more comments / or a paragraph at the top to discuss the purpose and usage of the `watcher_example` notebook. Are the individuals intended to use the watcher script already going to know _why_ they would need to use it?

ngreenwald (changes requested):
Nice start, see below

toffy/fov_watcher.py
```diff
@@ -0,0 +1,239 @@
...
+            run_metadata = json.load(f)
+
+        # parse run_metadata and populate expected structure
+        for fov in run_metadata.get('fovs', ()):
+            run_order = fov.get('runOrder', -1)
+            scan = fov.get('scanCount', -1)
+            if run_order * scan < 0:
+                raise KeyError(f"Could not locate keys in {run_folder}.json")
+
+            fov_name = f'fov-{run_order}-scan-{scan}'
```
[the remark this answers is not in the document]
ackagel: essentially yes. I wasn't sure if there'd be a situation where an fov is skipped within the json or scan ever isn't one, so I decided to fill the 'completion' dictionary only with what's explicitly detailed in the json.  i'm open to fixing the ending to 'scan-1' if that's the only relevant case.

toffy/fov_watcher.py
```diff
@@ -0,0 +1,239 @@
...
+        """ initializes RunStructure by parsing run json within provided run folder
+
+        Args:
+            run_folder (str):
+                path to run folder
+            timeout (int):
+                number of seconds to wait for non-null filesize before raising an error
+        """
+        self.timeout = timeout
+        self.completion = {}
```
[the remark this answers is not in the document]
ackagel: sure. maybe `fov_progress`

toffy/fov_watcher.py
```diff
@@ -0,0 +1,239 @@
...
+            TimeoutError
+
+        Returns:
+            (bool, str):
+                whether or not both json and bin files exist, as well as the name of the point
+        """
+
+        # TODO: check watchdog path depth
+
+        fov_name, extension = path.split('.')[0:2]
```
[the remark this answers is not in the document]
ackagel: this kind of ties into the `FileCreatedEvent` q so I'll answer both here.  Basically, the `on_created` function is called whenever **any** file is added to the watched directory/subdirectories.  So, each file that's copied from the instrument will call `on_created` which will directly pass that path to this function here, which will update the completion/progress dictionary if it has the relevant keys.

toffy/fov_watcher.py
```diff
@@ -0,0 +1,239 @@
...
+        wait_time = 0
+        if fov_name in self.completion:
+            if extension in self.completion[fov_name]:
+                while os.path.getsize(path) == 0:
```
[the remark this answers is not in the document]
ackagel: because of the way the FileCreatedEvent works, that's an unreachable case (assuming no instant creation/deletion shenanigans).  I should note that with an error type though, so I'll add that.

toffy/test_utils.py
```diff
@@ -77,3 +82,59 @@ def generate_sample_fovs_list(fov_coords, fov_names, fov_sizes):
         )
 
     return sample_fovs_list
+
+
+# generation parameters for the extraction/qc callback build
+# this should be limited to the panel, foldernames, and kwargs
+class ExtractionQCGenerationCases:
```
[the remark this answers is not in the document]
ackagel: The global extraction uses the masses detailed in the fov's json. I didn't know if the plan was to have csv panels cary with the data as before, or 'embed' them in the json pre-acquisition. If we intent to enforce the former practice, then I can hardcode out 'default global-panel acceptance', and have it be a manual toggle as a kwarg or something.

toffy/watcher_callbacks.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Union, Tuple, Callable
+
+import pandas as pd
+
+from mibi_bin_tools.bin_files import extract_bin_files
+
+from toffy.qc_comp import compute_qc_metrics
+
+
+def build_extract_and_compute_qc_callback(panel: Union[Tuple[float, float], pd.DataFrame],
```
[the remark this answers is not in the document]
ackagel: Started the callback when the `compute_qc_metrcis` function took post-extraction tiffs as inputs, when it made more sense to extract and then compute qc metrics in the same call.  Merged in same changes a commit or two ago, saw that the function now takes bin_file_paths and quickly refactored it in place, but they can definitely be separate.

ackagel:
@srivarra Good call.  I'm planning on adding some more verbose usage instructions once everything's functional, i.e probably will add that for the next review cycle.

toffy/watcher_callbacks.py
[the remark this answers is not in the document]
ngreenwald: Yeah, I think having one callback per function makes sense. Once we add a couple more, I think it'll make sense to refactor this back to your original plan, where the tifs are extracted first and then passed to the functions, but for now we can leave it.

toffy/fov_watcher.py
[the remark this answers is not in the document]
ackagel: to clarify, this is the run json, so it would need every file name.  it would still make sense for it to store that but I guess someone at ionpath really loves re-implementing manual filepath construction everytime they interact with the json 🤷

toffy/fov_watcher.py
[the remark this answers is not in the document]
ackagel: that's my assumption. the only time this becomes an issue is in the case of something like `fov-1-scan-1.bin.evilfileextension` which I suppose could happen? I know Matlab likes doing this with *.m files; creating a copy of one called something like *.m._ or something? 

Anyways, I could validate that the length of `path.split('.')` is actually two, just in case.  That way it can throw a warning or something

ngreenwald:
The qc_metrics function expects the run folder to already exist. However, the watcher function doesn't create it before starting fov-level callbacks, which leads to the watcher crashing with a bad filepath error

ngreenwald:
The print statements get printed for each callback, rather than for each fov

image

toffy/fov_watcher_test.py
```diff
@@ -0,0 +1,101 @@
...
+        ]
+
+        # callbacks are not performed for skipped fovs
+        if add_blank:
+            fovs = fovs[1:]
+
+        for i, validator in enumerate(validators):
+            validator(os.path.join(tmpdir, f'cb_{i}', RUN_DIR_NAME), fovs)
+
+    pass
```
srivarra, line 101: Does this `pass` statement do anything?

ackagel: nope; I set up function autocomplete to add a `pass` to the end of new, blank function declarations to keep the active linter from getting mad at me haha.  I'll trim pass statements in the next feature branch (issue/design doc incoming)

srivarra: That's a nice feature.

ackagel:
this is ready to merge in, just needs approval @ngreenwald . callback 'chaining' design doc should be up in a bit, and that implementation will go in a new branch off of main

ngreenwald (approved):
Looks good

PR #60, opened 2022-03-19
Write co-registration settings to JSON param file

### LINKED ISSUE #54: Save new co-registration params to temp file rather than settings.py

Currently, `update_coregistration_params.ipynb` appends the new co-registration variables to `settings.py` each time. However, because this is a tracked file, it will cause issues when updating the repo.

Write updated co-registration params to a temp file instead.

In notebooks that need these params, If that temp file doesn't exist, read the co-registration params from `settings.py`. If it does, then we read from the temp file instead.

alex-l-kong:

Closes #54. Currently, after running `update_coregistration_params.ipynb`, the new conversion factors are appended directly to `settings.py`. This causes a lot of clutter and causes `settings.py` to diverge from `main`. We should write the params to untracked temp `.json` file instead.

The following changes have been added:

- `write_coreg_params` has been nuked: instead we save to a file called `stage_optical_coreg_params.json` in `toffy` directly in `update_coregistration_params.ipynb`.
- `convert_stage_to_optical` now includes a a default argument `stage_optical_coreg_path` set to `stage_optical_coreg_params.json`. If this path exists, then we read the coregistration conversion factors directly from the created `.json` file. Otherwise, we use the default factors in `settings.py`.


templates/update_coregistration_params.ipynb
```diff
@@ -136,14 +137,12 @@
...
+   "id": "f126bc6a-4695-4722-9d73-b9dc66bb7195",
    "metadata": {},
    "outputs": [],
    "source": [
-    "tiling_utils.write_coreg_params(\n",
-    "    coreg_params,\n",
-    "    fiducial_info['name']\n",
-    ")"
+    "with open(os.path.join('..', 'toffy', 'stage_optical_coreg_params.json'), 'w') as cp:\n",
+    "    json.dump(coreg_params, cp)"
```
alex-l-kong, line 145: Currently, this file name, `'stage_optical_coreg_params.json'`, is the default setting. Should we allow the user to specify different names or keep this constant?

ngreenwald (changes requested):
Okay, this is a good start. 

I think we should standardize the coreg file naming, and not give the user the option to modify it. In addition, lets completely remove the coreg setting defaults from settings.py. If there's no coreg file detected, the fucntions that rely on one should give the user an error message and direct them to run the notebook. The defaults are not going to transfer across instruments, so I think it makes more sense to force the user to run it once on each instrument. 

If we have a structured name scheme (i.e. coreg_params_yymmdd), then we can search for the files in the expected places. 

The alternative is to append successive coreg updates to the same file (coreg_params.json) with some text in the JSON about when they were added. Whichever you prefer.

alex-l-kong:
@ngreenwald I think it's easier to write successive coreg updates to the same file. We'll store each different co-registration parameter set in list format such as:

```
{
    'coreg_params': [
        {
            'STAGE_TO_OPTICAL_X_MULTIPLIER': ...,
            'STAGE_TO_OPTICAL_X_OFFSET': ...,
            ...
            'date': '220322'
        },
        {
            ...
            'date': '220323'
        }
    ]
}
```

In this way, we have one JSON file (as opposed to multiple that would create more clutter in `toffy` and be harder to keep track of) and simply take the most recent set of params created (the last element in `'coreg_params'`).

This will allow us to remove the `STAGE_TO_OPTICAL_...` params from `settings.py`. However, we'll still need to keep the `MICRON_TO_STAGE_...` constants around somewhere. I think it's easiest to leave these in `settings.py`...alternatively, we could put these in a separate `.json` file as well. I'm open to other ideas here.

alex-l-kong:
Per our meeting, the following have been updated:

1. `settings.py` no longer contains `STAGE_TO_OPTICAL_` constants, however we'll leave the `MICRON_TO_STAGE_` constants there
2. The `STAGE_TO_OPTICAL_` constants generated by `update_coregistration_params.ipynb` will now be saved to `toffy/coreg_params.json`. It will contain a key called `coreg_params` which defines a list of every set of conversion factors generated by the user. Each set of co-registraion params will also contain a key called `'date'` which will indicate the timestamp they were generated. Only the most recent (the last one appended to the `coreg_params` list) is used.
3. If the user has not run `update_coregistration_params.ipynb` before `autolabel_tma_cores.ipynb`, an error message will direct them to run the co-registration script first.

toffy/tiling_utils_test.py
```diff
@@ -897,6 +903,110 @@ def test_generate_validation_annot(check_dist, check_duplicates, check_mismatche
     assert generated_annot == actual_annot
 
 
+# NOTE: this only tests if the visualization runs wti hvalid parameters
```
ngreenwald, line 906: typo

ngreenwald (changes requested):
Looks good, just a typo

ngreenwald (approved):

PR #63, opened 2022-03-24
Polish watcher

ackagel:
adds more markdown descriptions to watcher notebook and adds simple watcher test.

this is in a new PR since the other one is getting bulky.

> _alternatively, a version of the watcher for anyone from Warsaw_

### DIFF
--- a/templates/example_watcher.ipynb
+++ b/templates/example_watcher.ipynb
@@ -22,9 +22,10 @@
    "outputs": [],
    "source": [
     "import os\n",
+    "import pandas as pd\n",
     "\n",
     "from toffy.fov_watcher import start_watcher\n",
-    "from toffy.watcher_callbacks import build_extract_and_compute_qc_callback"
+    "from toffy.watcher_callbacks import build_extract_callback, build_qc_callback"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


ackagel:
there are a lot of extraneous print's in the watcher script.  I think some of them could stay in, since it's otherwise pretty hard to follow the watcher's progress.  maybe if we really want some polish, we could add a progress bar in the future.  open to ideas and thoughts on the superfluity of print statements

templates/example_watcher.ipynb
```diff
@@ -1,15 +1,31 @@
 {
```
ngreenwald, line 1: <p>Not sure if it's worth adding readthedocs just for this, but we should at least have a comment here with the list of available callbacks and their arguments</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/63/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

toffy/fov_watcher.py
```diff
@@ -160,6 +167,8 @@ def on_created(self, event: FileCreatedEvent):
         """
         super().on_created(event)
 
+        print(f'{event.src_path} was just created...')
```
ngreenwald, line 170: I think one print message per FOV to say that it was identified/is being processed is enough. All of the callback-specific print statements are useful for debugging, but not needed while running imo

toffy/fov_watcher.py
```diff
@@ -179,16 +188,21 @@ def on_created(self, event: FileCreatedEvent):
...
             # run per_fov callbacks
             for fov_func in self.per_fov:
                 logf.write(
                     f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- '
-                    f'Running {fov_func.__name__} on {point_name}'
+                    f'Running {fov_func.__name__} on {point_name}\n'
                 )
-                fov_func(self.run_folder, point_name, self.watcher_out)
+
+                callback_dir = os.path.join(self.watcher_out, fov_func.__name__.split('_')[0])
```
ngreenwald, line 201: I think we'll want the option for different base paths for the different callbacks.

ngreenwald (changes requested):

toffy/fov_watcher.py
ackagel: Sure.  Right now, if you pass in `D:\\some_place` into `start_watcher`, it will create the subfolder `D:\\some_place\\run_name_XXXX` and put each callback's output into it a respective subfolder within there. In these terms do you want `D:\\some_place` to be configurable 'per-callback'?

ngreenwald (approved):

PR #66, opened 2022-03-28
Add ROI name prefix to tiled region FOV names

alex-l-kong:

In preparation for the tiled region interactive overlay, we need to address the underlying names of the FOVs generated for each tiled region. Right now, they are all in the form `R{m}C{n}`, however this means that `R1C1` will be indistinguishable from ROI1, ROI2, etc. Adding the ROI prefix will help disambiguate these (ex. if the ROI name is `region1`, then the FOVs generated will be named `region1_R1C1`, `region1_R2C1`, etc.).

Additionally, we need to standardize the use of `roi` as a variable identifier throughout the tiled regions functions. Right now, we use `fov` pretty generously. However, it will be easier to follow if we use `roi`.

See above.

### DIFF
--- a/toffy/settings.py
+++ b/toffy/settings.py
@@ -1,6 +1,6 @@
 # tiled regions
-REGION_PARAM_FIELDS = ['region_start_row', 'region_start_col', 'fov_num_row', 'fov_num_col',
-                       'row_fov_size', 'col_fov_size', 'region_rand']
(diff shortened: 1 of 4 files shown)


alex-l-kong:
On second thought, I don't think it's worth forcing `region` to `roi`. However, any instance where `fov` should be `region` will still need to be fixed.

ngreenwald (approved):
Nice, looks good

PR #70, opened 2022-03-31
Improves watcher tests

ackagel:
## **Purpose**

Adds more coverage to watcher testing.  Also adds generalizable directory output validation.

## **Implementation**

Parametrizes standard cases over intensity extraction.  Uses functools.partial to help create per-case validators within the watcher case

## **Remaining issues**

per_run testing, but that's left to whoever first implements a per_run callback

### DIFF
--- a/toffy/fov_watcher_test.py
+++ b/toffy/fov_watcher_test.py
@@ -15,9 +15,10 @@ from toffy.test_utils import WatcherCases, RunStructureTestContext, RunStructure
 from toffy.fov_watcher import start_watcher
 
 TISSUE_DATA_PATH = os.path.join(Path(__file__).parent, 'data', 'tissue')
+RUN_DIR_NAME = 'run_XXX'
 
 
-def _slow_copy_sample_tissue_data(dest: str, delta: int = 10):
+def _slow_copy_sample_tissue_data(dest: str, delta: int = 10, one_blank: bool = False):
     """slowly copies files from ./data/tissue/
 
     Args:
@@ -29,9 +30,14 @@ def _slow_copy_sample_tissue_data(dest: str, delta: int = 10):
 
     print('copytime')
 
-    for tissue_file in os.listdir(TISSUE_DATA_PATH):
-        shutil.copy(os.path.join(TISSUE_DATA_PATH, tissue_file), dest)
+    for tissue_file in sorted(os.listdir(TISSUE_DATA_PATH)):
         time.sleep(delta)
+        if one_blank and '.bin' in tissue_file:
+            # create blank (0 size) file
+            open(os.path.join(dest, tissue_file), 'w').close()
+            one_blank = False
+        else:
+            shutil.copy(os.path.join(TISSUE_DATA_PATH, tissue_file), dest)
     print('copies done!')
 
 
@@ -55,12 +61,10 @@ def test_run_structure(run_json, expected_files):
 
 
 # TODO: add tests for per_run when per_run callbacks are created
+@pytest.mark.parametrize('add_blank', [False, True])
 @parametrize_with_cases('per_fov_partial, per_run, validators', cases=WatcherCases)
-def test_watcher(per_fov_partial, per_run, validators):
+def test_watcher(per_fov_partial, per_run, validators, add_blank):
     with tempfile.TemporaryDirectory() as tmpdir:
-
-        RUN_DIR_NAME = 'run_XXX'
-
         per_fov = []
         for i, func in enumerate(per_fov_partial):
             cb_dir = os.path.join(tmpdir, f'cb_{i}', RUN_DIR_NAME)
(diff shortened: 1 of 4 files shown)


ngreenwald (changes requested):
Right now, the testing is set up as a nice integration test for the entire workflow. However, we don't test the individual pieces of the `RunStructure` or FOVEventHandler class. 

If mocking the inputs to the FOVEventHandler is going to be really complicated we can discuss which parts of worth skipping, but having at least simple tests for some of the components that the `start_watcher` function relies on will make debugging much easier.

ngreenwald:
Right. I think the main reason to test FOVEventHandler directly would be for our own benefit, not to catch bugs for users. In general, the shorter the stacktrace, the easier to debug when you're making changes. 

In particular, I imagine in the future we'll be expanding FOVEventHandler to do some preprocessing of the data for inputs that are shared across multiple callbacks

srivarra:
Could you add argument types and descriptions for these parameters in the docstring?

https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/fov_watcher_test.py#L47

https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/test_utils.py#L124-L127

srivarra (changes requested):
Could you add argument types and descriptions for these parameters in the docstring?

https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/fov_watcher_test.py#L47

https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/test_utils.py#L124-L127

toffy/fov_watcher_test.py
```diff
@@ -40,13 +42,16 @@ def _slow_copy_sample_tissue_data(dest: str, delta: int = 10):
...
 
 
-@parametrize_with_cases('per_fov_partial, per_run', cases=WatcherCases)
-def test_watcher(per_fov_partial, per_run):
+# TODO: add tests for per_run when per_run callbacks are created
+@parametrize_with_cases('per_fov_partial, per_run, validators', cases=WatcherCases)
+def test_watcher(per_fov_partial, per_run, validators):
     with tempfile.TemporaryDirectory() as tmpdir:
 
+        RUN_DIR_NAME = 'run_XXX'
```
alex-l-kong, line 50: Is there a reason this is a constant? If so, can it be put outside the test function?

alex-l-kong (commented):
One minor comment, otherwise looks good for now.

ngreenwald:
The timeout error exception is not caught appropriately. We should add a test for this and fix it

image

ackagel:
> Could you add argument types and descriptions for these parameters in the docstring?
> 
> https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/fov_watcher_test.py#L47
> 
> https://github.com/angelolab/toffy/blob/525335af6ff17eeb57570c58481d86ff9c9a02db/toffy/test_utils.py#L124-L127

for `check_extraction_dir_structure`, for sure.

The other function is a test function, which we usually don't add a docstring to.  In this case, the only arguments it will be provided are filled by the `WatcherCases` class, in `test_utils.py`.

If the way the code-body is working seems unclear, I could make some variable names more descriptive, add some comments to the code body, or break some of it into a different function (particularly reusable portions).

toffy/fov_watcher_test.py
ackagel: That's a constant because I retyped 'run_XX' into one of the functions and got confused why everything was breaking for a solid 5 minutes lol.  It should probably stay a constant but I agree, it should be in global scope.

ackagel:
@ngreenwald adding the changes you suggested in PR #53 , then requesting review

toffy/fov_watcher_test.py
```diff
@@ -26,9 +30,14 @@ def _slow_copy_sample_tissue_data(dest: str, delta: int = 10):
 
     print('copytime')
```
alex-l-kong, line 31: Is there a reason we need this print statement here?

toffy/fov_watcher_test.py
```diff
@@ -26,9 +30,14 @@ def _slow_copy_sample_tissue_data(dest: str, delta: int = 10):
...
-        shutil.copy(os.path.join(TISSUE_DATA_PATH, tissue_file), dest)
+    for tissue_file in sorted(os.listdir(TISSUE_DATA_PATH)):
         time.sleep(delta)
+        if one_blank and '.bin' in tissue_file:
+            # create blank (0 size) file
+            open(os.path.join(dest, tissue_file), 'w').close()
+            one_blank = False
+        else:
+            shutil.copy(os.path.join(TISSUE_DATA_PATH, tissue_file), dest)
     print('copies done!')
```
alex-l-kong, line 41: Same with this one

ngreenwald (approved):
Nice, looks good.

alex-l-kong (approved):
Looks good, some minor comments about print statements.

toffy/fov_watcher_test.py
ackagel: oops, just some vestigial debugging code

PR #84, opened 2022-04-15
Add scheme for renaming TMA manual FOV duplicates

### LINKED ISSUE #83: TMA remapping fails with multiple input FOVs of the same name

**Describe the bug**
When the manual JSON file contains multiple FOVs with the same name, we would expect those to get flagged as duplicates and displayed in the counter. However, the notebook instead only displays the second instance of the duplicated name

**To Reproduce**
Create a JSON with identically named FOVs. 
Load into remapping noteobok

alex-l-kong:

Closes #83. There doesn't exist a check against a user specifying multiple manual FOVs with the same name. The interactive visualization assumes all names are unique, so in the case of duplicate FOVs, only the very last one of each is plotted.

The goal of this PR is twofold:

1. Clearly notify the user of duplicate FOVs
2. Ensure the interactive visualization plots all sets of duplicate FOVs in a manner that makes their locations easy to identify

Add a function called `rename_duplicate_fovs` in `tiling_utils` which takes in `manual_fov_regions` as a parameter. Each subsequent duplicate for a particular FOV name gets appended with `_duplicate{n}` (where `n` is the nth duplicate of a name found). For example, if there are 3 instances of `'R1C1'`, the 2nd one gets renamed `'R1C1_duplicate1'`, and the 3rd `'R1C1_duplicate2'`.

The call to `rename_duplicate_fovs` is placed directly after loading in `manual_fov_regions`. This ensures that the names are corrected prior to running it through the interactive visualization.

**Remaining issues**

The way the FOVs are renamed might need to be adjusted.


ngreenwald (approved):
Looks great, nice and simple

PR #85, opened 2022-04-15
Rename fov folders

### LINKED ISSUE #10: Rename FOVs
Add helper function to rename folders of FOVs based on TMA mapping script name

### LINKED ISSUE #72: Renaming FOV folders
*This is for internal use only; if you'd like to open an issue or request a new feature, please open a bug or enhancement issue*

**Section 1: Design details**

**Relevant background**

Closes [issue #10](https://github.com/angelolab/toffy/issues/10).
Currently, the FOV folders outputted from MIBI are labeled using a FOV number instead of the name of the specific FOV. (ex: fov-32-scan-1) 
We would like to renamed them to be more informative (ex: MoQC-scan-1).

**Design overview**

Create a helper function to rename the subdirectories using specific FOV names. The FOV names will be retrieved from the metadata in a JSON file, and there are two possible ways this can be done.
  1. Use the run JSON file and iterate through each FOV listed to retrieve the names. 
      - This would rely on the FOVs to be enumerated in the same order they are listed in the run file. 
  2. Retrieve each name from the individual JSON file for that specific FOV.

*Efficiency*
  1. Method 1 requires loading in a much larger file, but since there is a value `scanCount` we can use that to know how many duplicate FOV name folders there are and not have to reread the same FOV name multiple times.
  2. Method 2 uses smaller files, but it is probably less efficient to load every JSON file in the directory. Especially if there are multiple scans, since each scan produces its own folder and JSON file for the same FOV (ex: fov-32-scan-1, fov-32-scan-2, fov-32-scan-3).
     
**Design list/flowchart**

Assuming we can just use the run JSON file (Method 1):
* verify FOV folders are in the path specified
* load the JSON file
* iterate through each FOV listed
    - retrieve `scanCount` and `name`
* verify the name retrieved is valid
* rename the FOV folders accordingly 
    - verify all folder names were changed (the number of FOVs times `scanCount`)

**Required inputs**
The path to the folder containing all of the FOV subfolders (and the JSON file(s) needed). 

**Output files**

The directory names will be overwritten as stated above.

Once you have completed section 1, please tag the relevant parties and iterate on the initial design details until everyone is satisfied. Then, 
proceed to section 2

**Control flow**

Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

**Milestones and timeline**

List each of the major components of the project, and provide an estimated completion date for each one.

camisowers:

Closes #72.
Closes #10.
Renames the FOV subdirectories which are currently specified by both a FOV and scan number. Also gives the option so specify a new directory name to copy subdirectories to and edit those, leaving the originals intact.

In file **rename_fovs.py** I set up the function `rename_fov_dirs(run_path, fov_dir, new_dir=None)` which takes in a path to the run JSON file, the path to directory storing all the FOV subdirectories, and a new directory name (string) if desired. FOV name and scan count are read from the JSON file and stored as keys and values in `fov_scan`. The names of current subfolders are retrieved via `io_utils.listfolders()` and then sorted to ensure they're sorted alphabetically and numerically. An error is raised the expected number of folders based on the run file does not match the number of folders in the provided `fov_dir`. 

**Remaining issues**

I was hoping to have some kind of validation step ensuring all FOV names pulled were correct. Is there a master list of all possible FOVs to use? Still working on possible tests. Also will need some direction on where exactly to implement the function.

### DIFF
--- a/toffy/rename_fovs.py
+++ b/toffy/rename_fovs.py
@@ -75,10 +75,9 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
             new_name = os.path.join(change_dir, fov_scan[folder])
             os.rename(fov_subdir, new_name)
 
-
-'''
+"""
 #r = os.path.join("data", "json_test", "2022-04-07_TONIC_TMA21_run1.json")
 r = os.path.join("data", "json_test", "2022-01-14_postsweep_2.json")
 f = os.path.join("data", "fov_folders")
 n = os.path.join("data", "new_names")
-rename_fov_dirs(r, f, n)'''
+rename_fov_dirs(r, f, n)"""
--- a/toffy/rename_fovs.py
+++ b/toffy/rename_fovs.py
@@ -4,6 +4,28 @@ import warnings
 from distutils.dir_util import copy_tree
 
 from ark.utils import io_utils, misc_utils
+from toffy.tiling_utils import rename_duplicate_fovs
+
+
+def check_unnamed_fovs(fov_data):
+    """Identify FOVs that are missing the 'name' key and create one with value placeholder_{n}
+    Args:
+        fov_data (dict): the FOV run JSON
+
+    Returns:
+        dict: the same run JSON with placeholder names for FOVs that lack one
+       """
+
+    # count of FOVs that are missing the 'name' key
+    missing_count = 0
+
+    # iterate over each FOV and add a placeholder name if necessary
+    for fov in fov_data['fovs']:
+        if 'name' not in fov.keys():
+            missing_count += 1
+            fov_data[fov['name']] = f'placeholder_{missing_count}'
+
+    return fov_data
 
 
 def rename_fov_dirs(run_path, fov_dir, new_dir=None):
@@ -24,12 +46,14 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
     io_utils.validate_paths(run_path)
     io_utils.validate_paths(fov_dir)
 
-    # insert some kind of fov name validation
+    with open(run_path) as file:
+        run_metadata = json.load(file)
+
+    # check for missing or duplicate fov names
+    run_metadata = check_unnamed_fovs(run_metadata)
+    run_metadata = rename_duplicate_fovs(run_metadata)
 
     # retrieve custom names and number of scans for each fov, construct matching default names
-    with open(run_path) as f:
-        run_metadata = json.load(f)
-
     fov_scan = dict()
     for fov in run_metadata.get('fovs', ()):
         custom_name = fov.get('name')
@@ -75,9 +99,8 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
             new_name = os.path.join(change_dir, fov_scan[folder])
             os.rename(fov_subdir, new_name)
 
-"""
 #r = os.path.join("data", "json_test", "2022-04-07_TONIC_TMA21_run1.json")
-r = os.path.join("data", "json_test", "2022-01-14_postsweep_2.json")
-f = os.path.join("data", "fov_folders")
-n = os.path.join("data", "new_names")
-rename_fov_dirs(r, f, n)"""
+#r = os.path.join("data", "json_test", "2022-01-14_postsweep_2.json")
+#f = os.path.join("data", "fov_folders")
+#n = os.path.join("data", "new_names")
+#rename_fov_dirs(r, f, n)
--- a/toffy/rename_fovs.py
+++ b/toffy/rename_fovs.py
@@ -52,11 +52,10 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
     #testing
     #test = list(fov_scan)[0:5]
     #fov_scan = {t: fov_scan[t] for t in test}
-    #print(fov_scan)
 
     #check that fovs & scan counts match the number of existing FOV directories
-    if name_count != dir_count:
-        raise ValueError(f"FOV folders do not match expected amount listed in the run file")
+    #if name_count != dir_count:
+    #    raise ValueError(f"FOV folders do not match expected amount listed in the run file")
 
     # create new directory and copy contents of fov_dir
     if new_dir is not None:
@@ -71,8 +70,9 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
     for folder in fov_scan:
         fov_subdir = os.path.join(change_dir, folder)
         new_name = os.path.join(change_dir, fov_scan[folder])
-        os.rename(fov_subdir, new_name)
-        renamed_dirs += 1
+        if os.path.isdir(fov_subdir):
+            os.rename(fov_subdir, new_name)
+            renamed_dirs += 1
 
 
 
--- a/toffy/rename_fovs.py
+++ b/toffy/rename_fovs.py
@@ -47,6 +47,11 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
     io_utils.validate_paths(run_path)
     io_utils.validate_paths(fov_dir)
 
+    # check that new_dir doesn't already exist
+    if new_dir is not None:
+        if os.path.exists(new_dir):
+            raise ValueError(f"The new directory supplied already exists: {new_dir}")
+
     with open(run_path) as file:
         run_metadata = json.load(file)
 
@@ -87,7 +92,6 @@ def rename_fov_dirs(run_path, fov_dir, new_dir=None):
 
     # validate new_dir and copy contents of fov_dir
     if new_dir is not None:
-        io_utils.validate_paths(fov_dir)
         copy_tree(fov_dir, new_dir)
         change_dir = new_dir
     else:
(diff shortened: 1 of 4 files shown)


toffy/rename_fovs.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+
+    io_utils.validate_paths(run_path)
+    io_utils.validate_paths(fov_dir)
+
+    #retieve FOV names and number of scans for each
+    with open(run_path) as f:
+        data = json.load(f)
+        run_data = data['fovs']
+        fov_scan = {x['name']: x['scanCount'] for x in run_data}
```
ngreenwald, line 25: I think it will simplify the logic of the downstream steps if you construct this dictionary to instead be `{current_name: new_name}`. Then you can loop over each sub-folder, and use its current name to get the new name

ngreenwald: Here's the code adam used to construct the fov-x-scan-y names from the run json:
image

This will get you the current name of each FOV. For the new names, lets do the following: If there is only one scan of a given fov, then the name is just the user-supplied name (from the `name` field). Otherwise, if there is more than 1 scan, then scans 2-n are named `username-n`.

toffy/rename_fovs.py
```diff
@@ -0,0 +1,74 @@
...
+    dir_count = len(old_dirs)
+    old_dirs.sort()
+
+    #testing
+    #test = list(fov_scan)[0:5]
+    #fov_scan = {t: fov_scan[t] for t in test}
+    #print(fov_scan)
+
+    #check that fovs & scan counts match the number of existing FOV directories
+    if name_count != dir_count:
```
ngreenwald, line 46: If there are folders which are not in the JSON, lets raise an error. However, if there entries from the JSON which are not present as folders, then lets just raise a warning.

toffy/rename_fovs.py
```diff
@@ -0,0 +1,74 @@
...
+        raise ValueError(f"FOV folders do not match expected amount listed in the run file")
+
+    # create new directory and copy contents of fov_dir
+    if new_dir is not None:
```
ngreenwald, line 50: Instead, lets assume that the new_dir is the full path the new folder

toffy/rename_fovs.py
```diff
@@ -0,0 +1,74 @@
...
+        parent = os.path.join(fov_dir, os.pardir)
+        new_dir = os.path.join(parent, new_dir)
+        copy_tree(fov_dir, new_dir)
+
+    #change the FOV directory names
+    renamed_dirs = 0
+    for fov in fov_scan:
+        scan_num = fov_scan[fov]
+        for scan in range(1, scan_num+1):
+            if new_dir is not None:
```
ngreenwald, line 60: Because this will be true for all fovs in the JSON, this logic doesn't need to occur within the for loop.

ngreenwald (changes requested):
Great start! See below for comments. Also, notice that the "closes" text from the PR description isn't underlined in blue. This means github didn't understand what issue you wanted to close. Take a look at some of the other open PRs, then edit your description to match. You'll also want to close the original issue in addition to the design doc

toffy/rename_fovs.py
```diff
@@ -0,0 +1,80 @@
...
+            fov_scan[fov_name] = name
+
+    #insert some kind of fov name validation
+
+
+    #retrieve the current FOV directory names
+    old_dirs = io_utils.list_folders(fov_dir, "fov")
+
+    #check that fovs & scan counts match the number of existing FOV directories
+    if not set(old_dirs).issubset(set(fov_scan.keys())):
```
ngreenwald, line 54: See if you can use the existing [verify](https://github.com/angelolab/ark-analysis/blob/4fcd44ce4e1edbbbf75fbaab69867bbedb39d89a/ark/utils/misc_utils.py#L146) functions for some of these

ngreenwald (changes requested):
Looking nice, see below

toffy/rename_fovs.py
camisowers: verify_in_list() raises its own custom error printing the extra default_name directory that is not found in the json
Screenshot (230)
so do we want to raise ValueError(f"FOV folders exceed the expected amount specified in {run_path}") on top of that to be more specific? or just leave it as it

ngreenwald: If you can name the variables passed to `verify` such that the error message is informative enough, then using that logic would be preferable. Otherwise, can do custom

toffy/rename_fovs_test.py
```diff
@@ -0,0 +1,151 @@
...
+        os.mkdir(os.path.join(base_dir, 'not_new_directory'))
+        not_new_dir = os.path.join(base_dir, 'not_new_directory')
+
+        # existing directory for new_dir should raise an error
+        with pytest.raises(ValueError, match="already exists"):
+            rf.rename_fov_dirs(run_dir, fov_dir, not_new_dir)
+
+        # regular sample run data
+        ex_data = pd.DataFrame(
+            {'names': ['MoQC', 'tonsil_bottom', 'moly_qc_tissue'],
```
ngreenwald, line 95: I would have the custom names also follow a pattern, like custom_1, custom_2, etc. Otherwise it makes it seem like these names specifically mean something

ngreenwald, line 95: Having the names be more general to indicate what they are (run_1, run_2, run_3) is better than the specific names, since we're making them from scratch

toffy/rename_fovs_test.py
```diff
@@ -0,0 +1,151 @@
...
+        with pytest.raises(ValueError, match=r"already been renamed"):
+            rf.rename_fov_dirs(ex_run_path, fov_dir)
+        remove_fov_dirs(fov_dir)
+
+        # create correct fov folders
+        correct_fovs = ['fov-1-scan-1', 'fov-2-scan-1', 'fov-2-scan-2', 'fov-3-scan-1']
+        create_sample_fov_dirs(correct_fovs, fov_dir)
+
+        # test successful renaming to new dir
+        rf.rename_fov_dirs(ex_run_path, fov_dir, os.path.join(base_dir, 'new_directory'))
```
ngreenwald, line 124: Make sure to check that the correct names are there

ngreenwald, line 124: Make sure to check that the correct names are there

toffy/rename_fovs_test.py
```diff
@@ -0,0 +1,151 @@
...
+        remove_fov_dirs(os.path.join(base_dir, 'new_directory'))
+
+        # test successful renaming
+        rf.rename_fov_dirs(ex_run_path, fov_dir)
```
ngreenwald, line 128: Same here

ngreenwald, line 128: Same here

ngreenwald (changes requested):
Nice job, see below for comments

toffy/rename_fovs.py
```diff
@@ -0,0 +1,104 @@
...
+    # retrieve current default directory names, check if already renamed
+    old_dirs = io_utils.list_folders(fov_dir)
+    if set(old_dirs) == set(fov_scan.values()):
+        raise ValueError(f"All FOV folders in {fov_dir} have already been renamed")
+
+    # check if custom fov names & scan counts match the number of existing default directories
+    try:
+        verify_in_list(default=list(fov_scan.keys()), existing_folders=old_dirs)
+    except ValueError:
+        warnings.warn(f"Not all FOVs specified in {run_path} have an existing directory")
```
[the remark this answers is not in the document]
camisowers: Right now I have the ValueError being suppressed in favor of a warning, but if I remove the handling `verify_in_list` will output the first 10 default directory names (via `misc_utils.create_invalid_data_str`) that should've existed according to the run file.

Screenshot (234)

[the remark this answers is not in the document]
camisowers: Could edit the function to something like `def verify_in_list(warn=False,**kwargs):` ...
` if warn: warnings.warn(err_str)`
`else: raise ValueError(err_str)`

toffy/rename_fovs.py
```diff
@@ -0,0 +1,107 @@
...
+def rename_missing_fovs(fov_data):
+    """Identify FOVs that are missing the 'name' key and create one with value placeholder_{n}
+    Args:
+        fov_data (dict): the FOV run JSON data
+
+    Returns:
+        dict: a copy of the run JSON data with placeholder names for FOVs that lack one
+       """
+
+    copy_fov_data = fov_data
```
ngreenwald, line 20: Check to confirm if this actually makes a copy, or just a reference to the original. I can never remember with dicts what the default is.

ngreenwald (changes requested):
Looks great! See one comment below. 

Also, can you add the `rename_missing_fovs` function to the following functions in the toffy repo:

tiling_utils.set_tiled_region_params
tiling_utils.generate_tma_fov_list

As well as right before the rename duplicates call here in the tma notebook:
image

You can take a look at https://github.com/angelolab/toffy/blob/main/templates/2_create_tiled_mibi_run.ipynb
and https://github.com/angelolab/toffy/blob/main/templates/2_create_tma_mibi_run.ipynb
to see how those functions are called/used to figure out where the best place to put the helper function is

ngreenwald (approved):
Awesome job

PR #99, opened 2022-05-03
implements watcher chaining architecture

### LINKED ISSUE #89: Watcher Callback Chaining
### **Relevant background**

The watcher script takes a list of callback functions to be run on each FOV as it's discovered by the filesystem.  Currently, each function takes a `run_folder` and a `point_name` argument and performs all operations off of the bin file, which means the bin file may be needlessly re-extracted for each callback.  Further, this limits the scope of the callbacks to rely on the bin file data, as opposed tiff data, cell tables, etc.

A way to preserve relevant information and/or chain callback outputs resolves this issue.

### **Design overview**

The core component of this project is adding an information preserving and/or callback chaining system into the `start_watcher` system.  While there are a few implementation options, the user interface shouldn't be made any more complex, i.e do not require something akin to higher order function usage.

### **Design list/flowchart**

I see two recommendable implementation options here.  I prefer Option 2, but I figured I'd show both.

**Option 1: The Mega Function**

Essentially, have all callback functions exist as arguments/options in one large function.  Chaining/information preservation logic is explicitly written for relevant argument pairings.  Implementation pseudocode:

```
def build_mega_callback(extract_tiffs: bool, generate_qc: bool, do_something_with_qc_data: bool, **kwargs):
    def mega_callback(run_folder: str, point_name: str):
        bin_data = extract_bin(...)
        if extract_tiffs:
            make_tiffs(bin_data, **kwargs)
        if generate_qc:
            qc = make_qc(bin_data, **kwargs)
            if do_something_with_qc_data:
                something_with_qc(qc, **kwargs)
    
    return mega_callback
```

Pros:
 - most 'intention explicit'
 - minimal tweaks to existing codebase

Cons:
 - Probably doesn't scale with callbacks well (too much logic at some point)
 - Could create several different 'mega_callbacks' for different situations, but all would require `extract_tiffs` boilerplate
 - Tests probably scale even worse

**Option 2: "We're not cavemen Spongebob, we have Object Oriented Programming"**

This is similar to the mega function, but uses an object to preserve information.  This allows for more structured chaining logic.  This also makes multiple 'mega_callbacks' viable as the 'extract_tiffs' callback cab be made a base class for others to inherit from.  Implementation pseudocode:

```
# information preserving logic goes here
@dataclass
class FovCallbacks:
    run_folder: str
    point_name: str
    panel: pd.DataFrame = field(None, init=False)
    fov_data: xr.DataArray = field(None, init=False)
    
    def extract_tiffs(self, out_dir: str, panel: pd.DataFrame, **kwargs):
        self.panel = panel
        self.fov_data = extract_bin(...)
        ...
    
    def generate_qc(self, out_dir: str, panel: pd.DataFrame = None, **kwargs):
        if panel == self.panel or panel is None:
            # qc with self.fov_data
            ...
        else:
            # re-extract with new panel
            ...

# this callback building function will not need to be updated
# this is just one way of doing it. there are some cleaner ways that slightly alter user interface
def build_fov_callback(*args, **kwargs):
    def fov_callback(run_folder: str, point_name: str):
        callback_obj = FovCallbacks(run_folder, point_name)
        for arg in args:
            # walrus, python>=3.8
            if cb := getattr(callback_obj, arg, None):
                cb_kwargs = dict((k.split(f'{arg}_')[-1], kwargs[k]) for k in kwargs.keys() if arg in k)
                cb(**kwargs)

    return fov_callback        ​
```

Pros:
 - Very clean/segmented organization
 - Scales well and has intuitive ways of breaking apart related functionality
 - callback assembling logic is largely fixed; code updates (i.e new callbacks + chaining logic) go into class
 - Since functionality is well separated, testing can be well separated as well.

Cons:
 - Cleanest implementation might require some tweaks to the user interface. This would be a dictionary of FovCallback method names as keys, with sub dictionary of argname, argvalue pairs (i.e kwargs).  The example pseudocode uses the existing user interface and is clearly a bit messy.
 - The functional programmer hegemony is displeased by these mutable objects; an offering must be prepared to satiate them.

**Alternative Option**

Just to get it out of the way, there is another option.  We could edit RunStructure and/or FOV_EventHandler to stitch functions together at runtime.  There's several variations to this approach but they're all largely hard to maintain and essentially give RunStructure and/or FOV_EventHandler much more responsibility over data handling + state awareness than they really should ever have.

### **Required inputs**

Callback argument specifications.  This could be a list of method names + all kwargs, or a dictionary of method names to specific kwargs.

### **Output files**

No major change from existing implementation.  Possibly retrieve log file path from FOV_EventHandler and write to log file.

## **Section 2: Implementation details**

see above, lol

ackagel:
## **Purpose**

Implements watcher chaining architecture.  closes #89 

## **Changes**

As described in #89 .  Slightly changes notebook interface.

## **Remaining issues**

Per-run callbacks aren't in yet; that should probably go into a different PR.  There were also some changes to the watcher notebook in a different PR, so I might need to merge/resolve those differences before merging into main.


toffy/test_utils.py
```diff
@@ -300,6 +299,11 @@ def case_default(self, intensity):
             check_qc_dir_structure,
         ]
         return [
-            functools.partial(build_extract_callback, panel=panel, intensities=intensity),
-            functools.partial(build_qc_callback, panel=panel),
+            functools.partial(
+                build_fov_callback,
+                'extract_tiffs',
+                'generate_qc',
+                panel=panel,
```
ngreenwald, line 306: These functions could use some comments

toffy/watcher_callbacks.py
```diff
@@ -1,81 +1,135 @@
...
+
+        The data and the panel are then cached members of the FovCallbacks object
+
+        Args:
+            panel (pd.DataFrame):
+                Panel used for extraction
+            **kwargs (dict):
+                kwargs are passed to `extract_bin_files`
+        """
+        self.__fov_data = extract_bin_files(
```
ngreenwald, line 31: This would be more readable with argument names for the variables

toffy/watcher_callbacks.py
```diff
@@ -1,81 +1,135 @@
...
+        if arg not in methods:
+            raise ValueError(
+                f'{arg} is not a valid FovCallbacks member\n'
+                f'Accepted callbacks are {methods}'
+            )
+        argnames = inspect.getfullargspec(getattr(FovCallbacks, arg))[0]
+        for argname in argnames:
+            if argname not in kwargs and argname != 'self':
+                raise ValueError(
+                    f'Missing necessary keyword argument, {argname}'
```
ngreenwald, line 119: ... for the following function {arg}

toffy/watcher_callbacks.py
```diff
@@ -1,81 +1,135 @@
...
+                )
+
+    # constructuct actual callback
+    def fov_callback(run_folder: str, point_name: str):
+        callback_obj = FovCallbacks(run_folder, point_name)
+        print(args)
```
ngreenwald, line 125: leftover print statement

ngreenwald (changes requested):
Looking nice. For everyone else, please comment on anything that's confusing since Adam will be handing this off once he leaves

I'm gonna merge in the docs update now, then the notebook can be updated for new syntax and user friendliness

toffy/test_utils.py
ackagel: just the watcher cases or the whole suite of watcher/callback test utils?

ngreenwald: Mostly the cases, hard to interpret what's going on with the partial function building, especially looking forward for adding in additional callbacks

toffy/test_utils.py
```diff
@@ -289,6 +288,23 @@ def case_extrabin(self):
...
+    the mangement (or somehow be aware) of the temporary directory where the testing happens.
+
+    per_run callbacks are currently not partialed since none are implemented.  That being said,
+    they probably will undergo the same treatment once some are implemented.
+
+    Validation functions will check that the directory/files created by each callback are correct.
+    Some maybe partialed for convinience, since some arguments, like the panel, are created here
+    and might as well be premptively passed, so that only "run dependent" arguments need to be
+    passed to the validators (i.e out directory and point names).
+    """
```
ackagel, line 307: @ngreenwald lmk if this works for now. 

Tbh, `class TestCases:` docstrings should probably be required+standardized at some point, so that it's easier to read at a glance.  Maybe something like:
```
class TestCases:
    """Test cases for blah blah blah
    
    Long winded diatribe about why these test cases look they way they look

    Cases return:
        (return_types[0]):
            description for return[0]
        (return_types[1]):
            ...
    """
```

happy to change it to a format like this, but just wanted to get the info down in the meantime.

ngreenwald: Yeah what you have there is great

srivarra (commented):
What is being logged here? https://github.com/angelolab/toffy/blob/76ab82f0ad32f7a23ff91aaa49ca160a6304f612/toffy/fov_watcher.py#L189

`watcher_callbacks::build_fov_callback` returns the function `fov_callback` however `fov_callback` doesn't seem to modify or return anything. Could you clarify what it is doing?
https://github.com/angelolab/toffy/blob/cb995f355de2d6338ff265ebe4f31ba19b47037d/toffy/watcher_callbacks.py#L105

ackagel:
> What is being logged here?
> 
> https://github.com/angelolab/toffy/blob/76ab82f0ad32f7a23ff91aaa49ca160a6304f612/toffy/fov_watcher.py#L189
> 
> `watcher_callbacks::build_fov_callback` returns the function `fov_callback` however `fov_callback` doesn't seem to modify or return anything. Could you clarify what it is doing?
> 
> https://github.com/angelolab/toffy/blob/cb995f355de2d6338ff265ebe4f31ba19b47037d/toffy/watcher_callbacks.py#L105

@srivarra When the bin files + json files are transferred over from the instrument computer, there's a chance that the bin file is created as a 0 byte file, but doesn't never gets filled in with the actual data.  This logs that event.

As for `fov_callback`, I'll write up some documentation to make this more clear.  The gist of it, is that the user passes the names of the member functions of `FovCallbacks` as `*args`.  These functions can then be retrieved as `getattr(FovCallbacks(run_folder, point_name), member_function_name, None)`, with `None` being returned if `member_function_name` isn't found.  The `**kwargs` are passed as the arguments to these retrieved member functions.  Also, instead of using `*args`, I could set it up to take in a single list/iterable argument if that would be more clear.

toffy/watcher_callbacks.py
```diff
@@ -1,81 +1,157 @@
...
+    methods = [attr for attr in dir(FovCallbacks) if attr[0] != '_']
+
+    # validate user callback settings
+    for arg in args:
+        # check that `arg` is a method of FovCallbacks
+        if arg not in methods:
+            raise ValueError(
+                f'{arg} is not a valid FovCallbacks member\n'
+                f'Accepted callbacks are {methods}'
+            )
```
alex-l-kong, line 133: Would it be better to use something like `misc_utils.verify_in_list` to print a list of invalid `args` in the case of multiple, or do you think it's fine to just list the first offending one?

toffy/watcher_callbacks.py
```diff
@@ -1,81 +1,157 @@
...
+        # check that required (non-keyword) arguments for `arg` is present in passed `**kwargs`
+        argnames = inspect.getfullargspec(getattr(FovCallbacks, arg))[0]
+        for argname in argnames:
+            if argname not in kwargs and argname != 'self':
+                raise ValueError(
+                    f'Missing necessary keyword argument, {argname} for callback function {arg}...'
+                )
```
alex-l-kong, line 140: Same as above.

alex-l-kong (commented):
Just one really minor comment.

toffy/watcher_callbacks.py
ackagel: good idea, `verify_in_list` is probably better here.

ackagel:
I think everything's in.  Either way, this version is good to merge in.

ngreenwald (approved):
Looks good

PR #114, opened 2022-05-30
added moly_path option

### LINKED ISSUE #90: Don't require Moly points for tiled region and TMA FOV generation

The tiled region and TMA notebooks currently require a Moly point to be specified even if the user does not want to insert them. We should only require this to be specified if the user actually wants inserted.

More precise conditional checking (for example, using the `moly_region` and `moly_interval` keys in the `region_params` `dict` for tiled regions).

srivarra:

Closes #90.

`moly_path` is now optional in `tiling_utils::generate_tiled_region_fov_list`.

Had to add a check for `tiling_params["moly_region"]`, and change several tests to work with the check.

**Remaining issues**

I was not able to find a way to keep the tests the same with this added condition.

### DIFF
--- a/toffy/tiling_utils.py
+++ b/toffy/tiling_utils.py
@@ -378,29 +378,29 @@ def set_tiled_region_params(region_corners_path):
     # store the read in parameters in the region_params key
     tiling_params['region_params'] = generate_region_info(region_params)
 
-    # whether to insert moly points between regions
-    moly_region_insert = read_tiling_param(
-        "Insert a moly point between each tiled region? Y/N: ",
-        "Error: moly point region parameter must be either Y or N",
-        lambda mri: mri in ['Y', 'N', 'y', 'n'],
-        dtype=str
-    )
+    # # whether to insert moly points between regions
+    # moly_region_insert = read_tiling_param(
+    #     "Insert a moly point between each tiled region? Y/N: ",
+    #     "Error: moly point region parameter must be either Y or N",
+    #     lambda mri: mri in ['Y', 'N', 'y', 'n'],
+    #     dtype=str
+    # )
 
-    # convert to uppercase to standardize
-    moly_region_insert = moly_region_insert.upper()
-    tiling_params['moly_region'] = moly_region_insert
+    # # convert to uppercase to standardize
+    # moly_region_insert = moly_region_insert.upper()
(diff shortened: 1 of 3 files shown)


toffy/tiling_utils.py
```diff
@@ -502,22 +503,25 @@ def generate_tiled_region_fov_list(tiling_params, moly_path):
...
             Data containing information about each FOV
     """
 
     # file path validation
-    if not os.path.exists(moly_path):
-        raise FileNotFoundError("Moly point file %s does not exist" % moly_path)
+    if (tiling_params.get("moly_region", "N") == "Y") or \
+       (tiling_params.get("moly_interval", 0) > 0):
+        if not os.path.exists(moly_path):
+            raise FileNotFoundError("Moly point file %s does not exist" % moly_path)
```
ngreenwald, line 520:
```suggestion
            raise FileNotFoundError("The provided Moly FOV file %s does not exist. If you want to include Moly FOVs you must provide a valid path. Otherwise, select 'No' for the options relating to Moly FOVs" % moly_path)
```

ngreenwald (changes requested):
One change about the error message to be more informative.

toffy/tiling_utils.py
```diff
@@ -377,29 +378,29 @@ def set_tiled_region_params(region_corners_path):
...
-        "Enter the FOV interval size to insert Moly points. If yes, enter the number of FOVs "
-        "between each Moly point. If no, enter 0: ",
-        "Error: moly interval must be 0 or a positive integer",
-        lambda mi: mi >= 0,
-        dtype=int
-    )
-
-    if moly_interval > 0:
-        tiling_params['moly_interval'] = moly_interval
+    # # whether to insert moly points between regions
```
ngreenwald, line 381: I talked to Hadeesha, he wants to leave this in. Can you update the user prompt for each of these to indicate that they'll need to provide the path to the example moly_FOV json file if they want this option?

ngreenwald (changes requested):
Hadeesha wants us to keep this as an option

toffy/tiling_utils.py
srivarra: yup yup

ngreenwald (approved):

PR #125, opened 2022-06-07
Mph norm workflow

### LINKED ISSUE #37: Image normalization todos
- Create a jupyter notebook to walk through the entire process
- Add logging to record normalization factor applied to channels in each fov
- Add functions to compute to MPH and counts for FOVs in a directory
- add global variable to control which masses are used across pipeline
- update paths in rosetta notebook for default toffy locations
- update readme

ngreenwald:

1. Adds functionality to the set_up_toffy notebook which allows the user to construct a tuning curve for their instrument. This takes in a detector sweep, and produces a tuning curve relating MPH to signal intensity
2. Creates the `4b_normalize_image_data` notebook which walks the user through the normalization process. This assumes the user has already constructed a tuning curve, and will allow them to easily normalize their data
3. Switches the logic for normalization. Previously, we tried to fit a relationship between mass and MPH within an FOV, which would then allow us to use a small number of MPH values from selected masses to normalize. However, this created more problems than it solved. Instead, we now compute the MPH for every mass in the panel. This is more compute intensive, but it is quite fast and won't be a major bottleneck. We then fit a separate curve for each mass over the course of the run, modeling how the MPH is changing for that mass as a function of run length. This results in very smooth estimates of MPH for each marker in the panel. We then use those adjusted mass-specific MPH values to normalize the data. 
4. Adds plotting and QC outputs to allow the user to visualize the curve fit, as well as get notified for any normalization values which are outside a proscribed range
5. Updates the 4a_rosetta notebook to have the same default paths as the rest of the notebooks. Also removes the nested for loop structure in favor of processing a single run, simplifying the code

Closes #37


toffy/normalize.py
```diff
@@ -184,69 +246,311 @@ def combine_tuning_curve_metrics(dir_list):
...
+    # loop over each fov
+    for fov in img_fovs:
+        # compute per-mass normalization constant
+        pulse_height_fov = pulse_height_df.loc[pulse_height_df['fov'] == fov, :]
+        norm_vals = norm_func(pulse_height_fov['pulse_height_fit'].values)
+        channels = pulse_height_fov['Target'].values
+
+        # get images
+        images = load_utils.load_imgs_from_tree(img_dir, fovs=[fov], channels=channels,
+                                                dtype='float32', img_sub_folder=img_sub_folder)
```
ackagel, line 552: Depending on how fast/slow this is, we might want to batch load the fovs, and then loop over each batch.

ackagel (commented):
Looks good so far.  It looks like a lot of my review comments in review-nb are in the works in `normalize.py`

templates/1_set_up_toffy.ipynb
```diff
@@ -15,9 +15,10 @@
    "id": "e36293c5-aa89-4029-a3fa-e8ea841bb8b5",
```
ackagel, line 15: <p>Can't the sensitivity curve generation go in 4b, since it's only used there?  Is the idea that putting this here will encourage people to run a sweep before data aq?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

ackagel, line 15: <p>Can't this be done programmatically here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

ackagel, line 15: <p>Any reason why the <code>normalize</code> functions can't be combined into a single <code>normalize.calibrate_from_sweep(sweep_path, moly_masses, ...)</code>wrapper function?  This could also handle the <code>norm_func.json</code> saving as well.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

templates/4b_normalize_image_data.ipynb
```diff
@@ -0,0 +1,142 @@
+{
```
ackagel, line 1: <p>Just to be explicit, I think having a <code>norm_json_path</code> variable passed here would re-enforce the sweep normalization step in the 1a notebook; it isn't immediately clear where the normalization parameters are coming from otherwise.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

templates/1_set_up_toffy.ipynb
alex-l-kong, line 15: <pre class="ql-syntax" spellcheck="false">Line #10.            normalize.write_mph_per_mass(base_dir=fov_path, output_dir=fov_path, fov='fov-1-scan-1', masses=moly_masses)
</pre><p>Would be aesthetically better to reduce the tab size to 4 spaces. Same for the line below.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

templates/4b_normalize_image_data.ipynb
alex-l-kong, line 1: <p>I'd recommend splitting this cell up into two cells:</p><p> </p><ol><li>Making the directories to hold the normalized images and associated processing files.</li><li>The loop over each FOV to generate the corresponding MPH values.</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/125/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

templates/1_set_up_toffy.ipynb
ngreenwald: <p>No, it's because it only needs to happen once per instrument, not separately for each run. So having it here means it won't be present in the notebook each time people are normalizing</p>

ngreenwald: <p>Sweeps aren't created in their own folder, they're just put into the main /Data folder. So you need to separately identify each FOV from the sweep, which are given the generic names. This was initially what that <code>find_detector_sweeps</code> function was for, but then Erin had a couple sweeps where FOVs were missing, so it gave an error. </p><p> </p><p>We could change it so that it would give a warning when an FOV is missing, rather than an error, and then ask people to list the first FOV and last FOV of their sweep and it would find the rest, but at that point it started to feel like the solution was almost as complicated as the problem. Up to you though, it would be an easy change </p>

ngreenwald: <p>Yeah agreed</p>

templates/4b_normalize_image_data.ipynb
ngreenwald: Added a description at the top of the notebook instead

PR #133, opened 2022-06-17
developing json dump and load wrappers

### LINKED ISSUE #129: Add helper function for reading and writing json files
We have a couple places in the codebase where we reuse the sample context manager for reading and writing jsons.

We should create dedicated helper functions that peform these actions, as well as sanity checking the paths.

mheskett:

Add file path validation to reading and writing individual json files.

https://github.com/angelolab/toffy/issues/129
closes #129 

Added a function to read and write JSONs that checks for a valid path using io_utils.validate_paths()
**Remaining issues**

If there are specific areas of your implementation that you feel need specific feedback, or if there are things you aren't sure about, please highlight them here

### DIFF
--- a/toffy/json_utils.py
+++ b/toffy/json_utils.py
@@ -80,3 +80,40 @@ def list_moly_fovs(bin_file_dir):
             moly_fovs.append(moly_name)
 
     return moly_fovs
+
+def read_json_file(json_path):
+    """Reads json file and returns json file object while verifying dirs exist
+
+    Args:
+        json_path (str): path to json file
+
+    Returns:
+        json file object"""
+
+    # call to validate paths will raise errors if anything wrong, and do nothing if 
+    # file path valid
+    io_utils.validate_paths(json_path)
+
+    with open(json_path, 'r') as jp:
+        json_file = json.load(jp)
+
+    return json_file
+
+def write_json_file(json_path,json_object):
+    """Writes json file object to json file. Raises error if directory doesnt exist.
+
+    Args:
+        json_path: full path to write json file
+    Returns:
+        nothing"""
+    
+    # get the path minus the proposed file name.
+    dir_path = os.path.dirname(os.path.abspath(json_path))
+
+    # Raises error if path doesnt exist
+    io_utils.validate_paths(dir_path)
+
+    with open(json_path, 'w') as jp:
+        json.dump(json_object, jp)
+
+    return
--- a/toffy/json_utils.py
+++ b/toffy/json_utils.py
@@ -82,7 +82,7 @@ def list_moly_fovs(bin_file_dir):
     return moly_fovs
 
 
-def read_json_file(json_path,encoding=None):
+def read_json_file(json_path, encoding=None):
     """Reads json file and returns json file object while verifying dirs exist
 
     Args:
@@ -95,13 +95,13 @@ def read_json_file(json_path,encoding=None):
     # file path valid
     io_utils.validate_paths(json_path, data_prefix=False)
 
-    with open(json_path, mode='r',encoding=encoding) as jp:
+    with open(json_path, mode='r', encoding=encoding) as jp:
         json_file = json.load(jp)
 
     return json_file
 
 
-def write_json_file(json_path, json_object,encoding=None):
+def write_json_file(json_path, json_object, encoding=None):
     """Writes json file object to json file. Raises error if directory doesnt exist.
 
     Args:
@@ -115,5 +115,5 @@ def write_json_file(json_path, json_object,encoding=None):
     # Raises error if path doesnt exist
     io_utils.validate_paths(dir_path, data_prefix=False)
 
-    with open(json_path, mode='w',encoding=encoding) as jp:
+    with open(json_path, mode='w', encoding=encoding) as jp:
         json.dump(json_object, jp)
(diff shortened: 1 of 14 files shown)


toffy/json_utils.py
```diff
@@ -80,3 +80,40 @@ def list_moly_fovs(bin_file_dir):
...
+    # get the path minus the proposed file name.
+    dir_path = os.path.dirname(os.path.abspath(json_path))
+
+    # Raises error if path doesnt exist
+    io_utils.validate_paths(dir_path)
+
+    with open(json_path, 'w') as jp:
+        json.dump(json_object, jp)
+
+    return
```
ngreenwald, line 119: Don't need a return statement if you're not going to return anything

toffy/json_utils_test.py
```diff
@@ -77,3 +77,28 @@ def test_list_moly_fovs(tmpdir):
     pred_moly_fovs = json_utils.list_moly_fovs(tmpdir)
 
     assert np.array_equal(pred_moly_fovs.sort(), moly_fovs.sort())
+
+
+def test_read_json_file():
```
ngreenwald, line 82: You should use a temporary directory here, otherwise the files get written to your actual computer. The temporary directory that gets created is treated like a new base directory. Then you can add folders, files, etc inside like normal

https://github.com/angelolab/toffy/blob/main/toffy/file_hash_test.py#L12

toffy/json_utils_test.py
```diff
@@ -77,3 +77,28 @@ def test_list_moly_fovs(tmpdir):
...
+    moly_json = {'name': 'bob',
+                 'standardTarget': 'Molybdenum Foil'}
+    json_path = cwd+"/test.json",
+    
+    # write test json
+    with open(json_path, 'w') as jp:
+        json.dump(moly_json, jp)
+
+    # Make sure errors come up when the directory is bad
+     with pytest.raises(ValueError, match='files are the same length'):
```
ngreenwald, line 94: The match here should be the error message that actually gets raised. So what error is validate_paths going to raise for an invalid path? You can run the code without catching the error to check and see what the exact syntax is, then update it here

toffy/json_utils_test.py
```diff
@@ -77,3 +77,28 @@ def test_list_moly_fovs(tmpdir):
...
+            json_utils.read_json_file("/neasdf1246ljea/asdfje12ua3421ndsf/asdf.json")
+
+    ## Now read json with new function assuming path is legit
+    newfile_test = json_utils.read_json_file(json_path)
+    with open(path, 'r') as jp:
```
ngreenwald, line 99: instead of reading in `newfile_standard`, you should assert that `newfile_test` is the same as `moly_json`

ngreenwald (changes requested):
Nice start! Make sure to actually fill our the PR description at the top. Also, put what issue # this is closes, immediately preceded by "closes", so that github knows to automatically close it. You can check the other open PRs for an example of the description and close syntax.

toffy/json_utils_test.py
```diff
@@ -77,3 +78,44 @@ def test_list_moly_fovs(tmpdir):
...
+        # Make sure errors come up when the directory is bad. Doesnt check to see
+        # if filename itself exists somewhere else bc that would involve recursion into directories
+        # This relies upon io_utils.validate_paths() being fully functional and tested
+        bad_path = "/neasdf1246ljea/asdfje12ua3421ndsf/asdf.json"
+        with pytest.raises(ValueError, match=f'The path, {bad_path}, is not prefixed with \'../data\'.\n'
+                    f'Be sure to add all images/files/data to the \'data\' folder, '
+                    f'and to reference as \'../data/path_to_data/myfile.tif\''):
+                json_utils.read_json_file(bad_path)
+
+        # Create bad dir path to invoke all 3 possibilities of validate_paths()
```
ngreenwald, line 105: So the testing philosophy is that you don't need to test control flow paths in all helper functions that are called, only the ones that are newly added. So it's not necessary that all cases in `validate_paths` are tested, since that function has its own set of tests. Instead, you just need to test the case where the path is valid, and the case where the path isn't valid

toffy/json_utils_test.py
```diff
@@ -77,3 +78,44 @@ def test_list_moly_fovs(tmpdir):
...
+        with pytest.raises(ValueError, match=''):
+               json_utils.read_json_file(tmp_dir+"/dkwn4823hjf08371gjsdfasdfa20135ndjsa.json")
+
+        # Read json with read_json_file function assuming file path is good
+        newfile_test = json_utils.read_json_file(json_path)
+
+        # Make sure using the read_json_file leads to the same object as moly_json
+        assert newfile_test == moly_json
+
+    return
```
ngreenwald, line 121: Now you can add the same tests for `read_json`. One test case where it's valid, one test where it's not.

ngreenwald (changes requested):
Looking good, see comments

ngreenwald (changes requested):
Looks great. If you look at the bottom of this page, you can see that the build is still failing. That's just because of style issues, like indentation, spacing, etc. You can check for those locally by adding --pycodestyle at the end of the pytest command, or click on the red x on github to see them. If you're curious about the formatting requirements, you can look up pep8, which explains most of them.  Once you address those, everything will be ready. 

The last thing to do is go through the toffy repo and replace anywhere that we use the json read/write logic with the new functions you wrote.

PR #142, opened 2022-06-28
Implement json helper funcs

mheskett:

Implement read_json_file and write_json_file into many locations throughout toffy.

 removed
`with open(path,"r") as fp:
        json.load/dump(fp)`
and replaced with read/write_json_file() function calls

**Remaining issues**

### DIFF
--- a/templates/2_create_tiled_mibi_run.ipynb
+++ b/templates/2_create_tiled_mibi_run.ipynb
@@ -34,7 +34,8 @@
    "source": [
     "import json\n",
     "import os\n",
-    "\n",
+    "from toffy.json_utils import read_json_file\n",
+    "from toffy.json_utils import write_json_file\n",
     "from toffy import tiling_utils"
    ]
   },
@@ -140,8 +141,7 @@
    "outputs": [],
    "source": [
     "# write the tiled output\n",
-    "with open(tiled_region_fovs_path, 'w', encoding='utf-8') as trfp:\n",
-    "    json.dump(tiled_region_fovs, trfp)"
+    "write_json_file(json_path=tiled_region_fovs_path, json_object=tiled_region_fovs, encoding='utf-8')"
    ]
   }
  ],
@@ -161,7 +161,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.12"
+   "version": "3.8.13"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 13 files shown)


templates/2_create_tiled_mibi_run.ipynb
```diff
@@ -34,7 +34,8 @@
    "source": [
     "import json\n",
     "import os\n",
-    "\n",
+    "from toffy.json_utils import read_json_file\n",
```
ngreenwald, line 37: These can be condensed into one line:
``` from toffy.json_utils import read_json_file, write_json_file```

File: toffy/fov_watcher_test.py [hunk 2]
```diff
@@ -14,6 +14,8 @@
 from toffy.test_utils import WatcherCases, RunStructureTestContext, RunStructureCases
 from toffy.fov_watcher import start_watcher
 from toffy.watcher_callbacks import build_callbacks
+from toffy.json_utils import read_json_file
```
ngreenwald, line 17: There's a couple files where you import both but don't use both. Only need to import the functions that are actually used

toffy/normalize.py
```diff
@@ -19,6 +19,9 @@
 from mibi_bin_tools.bin_files import extract_bin_files, get_median_pulse_height
 from mibi_bin_tools.panel_utils import make_panel
 
+from toffy.json_utils import read_json_file
```
ngreenwald, line 22: Same thing here, and in the other files in this PR where you import both; cleaner to put on a single line

toffy/rosetta.py
```diff
@@ -26,8 +28,7 @@ def transform_compensation_json(json_path, comp_mat_path):
     returns:
         pd.DataTable: matrix with sources channels as rows and target channels as columns"""
 
-    with open(json_path) as f:
-        data = json.load(f)['Data']
+    data = read_json_file(json_path)['Data']
```
ngreenwald, line 31: Only one of the imports is used in this file, don't need to import both

ngreenwald (changes requested):
Looks great! Just some formatting changes

templates/2_create_tiled_mibi_run.ipynb
mheskett: question, whats the philosophy for spaces between import statements? i see you import some in groups and then have spaces

ngreenwald: We try and follow this guidance, though it doesn't alway stay up to date:
https://peps.python.org/pep-0008/#imports

PR #174, opened 2022-07-27
Simplify QC metric visualization procedure

### LINKED ISSUE #156: Default display just non-zero mean intensity chart in QC notebook

The QC notebook currently displays 3 metrics: non-zero mean intensity, total intensity, and 99.9% intensity value. This can be confusing to the user. In most cases, the only chart that is needed to determine anomalies in the extraction process is non-zero mean intensity.

instead of making 3 separate calls to each metric, we should default the visualization to non-zero mean intensity. However, we should also provide variables (with appropriate documentation) the user can change to visualize the other two stats.

**Additional context**

With the addition of the MPH notebook, the results of the QC notebook are less critical, but still nice to have around.

### LINKED ISSUE #162: Remove certain channels for QC metric visualization

While all of the channels are needed for computations such as rosetta, for QC metric visualization some of these will not be useful for the user. Channels like `Background`, `chan_39`, etc. take up a lot of space and should be removed.

Define a constant `CHANNEL_FILTER` that is used to filter out the channels to visualize.

alex-l-kong:

Closes #156. Closes #162. Users will likely just want to view `'Non-zero mean intensity'` stats, so we don't need to display the other two ('Total intensity', '99.9% intensity value'), instead providing the user an option to change the QC metric visualized.

Additionally, users will not need to view the QC stats for certain calibration or background channels (`'chan_'` prefixed channels, `'Au'`, `'Noodle'`).

Update the notebook and `visualize_qc_metrics` to ensure:

- Just one visualization cell is contained. A variable `qc_metric` is preset to `'Non-zero mean intensity'`, with a note that it can also be set to `'Total intensity'` or `'99.9% intensity value'`
- The user will not need to see the name of the underlying `combined_` file in the notebook. `visualize_qc_metrics` will generate the name of that file for the provided QC metric using the index positions in `QC_SUFFIXES` and `QC_COLUMNS` in `settings.py`
- A channel filter is included in `visualize_qc_metrics` that removes `'Au'`', `'Noodle'`, and any channel prefixed with `'chan_'`


templates/3c_generate_qc_metrics.ipynb
```diff
@@ -164,96 +164,34 @@
     "\n",
```
ngreenwald, line 164: <p>We should update this to include a run that has more of a decrease over time, to show what level of decrease is expected. That can be a separate PR to just update the image. Will you make an issue? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/174/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

ngreenwald (approved):

PR #194, opened 2022-08-03
Adjust min TMA size to 2x2

alex-l-kong:

Previously, the TMA script enforced a strict 3x3 minimum grid size. @HPiyadasa recently had an application that required a 2x11, which made it more difficult for him. Per a discussion with @ngreenwald, we can reduce the min grid size to 2x2.

Change the validation check of `num_fov_row` and `num_fov_col` in `generate_tmv_fov_list` to throw a `ValueError if `< 2` instead of `< 3`.

**Remaining issues**

Do we want to support grids that contain a dimension of 1 on either axis?

### DIFF
--- a/toffy/tiling_utils.py
+++ b/toffy/tiling_utils.py
@@ -664,12 +664,12 @@ def generate_tma_fov_list(tma_corners_path, num_fov_row, num_fov_col):
             "TMA corners file %s does not exist" % tma_corners_path
         )
 
-    # user needs to define at least 3 FOVs along the x- and y-axes
-    if num_fov_row < 3:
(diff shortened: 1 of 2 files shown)


ngreenwald (approved):

HPiyadasa:
Is there any purpose of having a min grid size at all? Maybe replace the
requirement with just a warning msg?

Hadeesha

ngreenwald:
I don't think there's a point in using the script for a single row/column of FOVs right?

HPiyadasa:
If there’s any staining gradient on the slide either across or running
down, it will be helpful to decouple it from any instrument issues that may
be influence the run.

PR #196, opened 2022-08-04
Ensures co-registration path is absolute

### LINKED ISSUE #184: Support more flexible file path specification for `coreg_params.json` saving

Currently, `1_set_up_toffy.ipynb` saves `coreg_params.json` to `os.path.joinI('..', 'toffy', 'coreg_params.json')`. This makes the assumption that the user is going to be working with notebooks directly in `templates`.

We've had several users make sub-directories inside `templates` on Drago during our test runs. This is to make it easier for them to locate their notebook copies and not confuse them with others. Unfortunately, putting the setup notebook in a directory means the preset path to `coreg_params.json` will not work.

We should allow the user to set a path to the desired location to save `coreg_params.json`. This can default to `os.path.joinI('..', 'toffy', 'coreg_params.json')` in the actual call to `save_coreg_params`, but it should also exist as a variable inside the setup notebook that the user can change.

This change will also need to be propagated to notebooks that rely on co-registration params, namely, the two tiling notebooks.

alex-l-kong:

Closes #184. Currently, saving `coreg_params.json` correctly assumes the user is working directly in the `templates` directory when starting up their Jupyter notebook. However, users and more frequently creating subdirectories inside `templates` to organize their runs, which causes this saving scheme to fail.

Using an absolute path to `coreg_params.json` in `toffy` should do the trick.


ngreenwald (approved):
The build passed, looks like it's not updating from travis for some reason

PR #209, opened 2022-08-13
Add checks for out-of-bounds coordinates during tiling process

### LINKED ISSUE #197: Address OOB issues with co-registration and tiling
**Section 1: Design details**

**Relevant background**

Follows up on #141.

The current co-registration process assumes the user inputs the correct values for stage and equivalent optical coordinates. However, this may not always be the case. `1_set_up_toffy.ipynb` should catch these ahead of time.

Additionally, for both tiling notebooks (`2_create_tiled_mibi_run.ipynb` and `2_create_tma_mibi_run.ipynb`), the check needs to extend to the FOVs created for the run file.

**Design overview**

Having verified that the fiducial conversion factors don't drastically differ between Drago and Creed, we can use those as a reference and verify that the generated conversion factors don't get assigned a value more than 20% from those baselines.

Slide boundary references also need to be added to ensure no FOV gets placed outside the slide. This will apply for 2 steps in the tiled region and TMA processes:

- During the verification of the top-left region corners for tiled regions, and the corner FOVs for TMAs
- Generating the FOVs within each tiled region/TMA

**Design list/flowchart**

Basic validation will be handled in the `1_set_up_toffy.ipynb` notebook during both the user input and co-registration param generation cells. If any coordinate or co-registration param falls out of the accepted range for their respective factor, we should report this as an error.

Validating the actual FOVs generated for each tiled region/TMA will be handled in their respective `2_` notebook. For tiled regions, this will happen during both the user input and FOV generation cells. For TMAs, this will happen only during the FOV generation cell (because this masks the call to validate the TMA corners).

**Required inputs**

No more than what's already required for each notebook.

**Output files**

N/A, just a few new error messages to display in case of OOB.

**Control flow**

`settings.py` will contain constants that define the hard-coded baseline constant conversion factors, as well as the left, right, top, and bottom boundaries of the slide in both stage and optical coordinates.

We'll define 2 helper functions:

1. `verify_coordinate_on_slide`: checks that a coordinate doesn't fall out of the acceptable slide boundary. This will take an argument `coord_type` which can be set to either `'stage'` or `'optical'`
2. `verify_coreg_param_tolerance`: checks that the co-registration parameters don't differ by more than 20% from its acceptable baseline

Both of these functions will throw a `ValueError` if verification fails.

The validating functions will be used by:

1. `generate_coreg_params`: the coordinate validator needs to be run for both stage and optical coordinate inputs. The co-registration parameter validator will verify the final conversion factors generated.
2. `read_tiled_region_inputs`: the coordinate validator runs on the `'centerPointMicrons'` provided for each corner FOV.
3. `generate_tiled_region_fov_list`: before adding a FOV to the list, run the coordinate validator runs on each FOVs `'centerPointMicrons'`.
4. `validate_tma_corners`: add calls to the coordinate validator for each corner
5. `generate_tma_fov_list`: before adding a FOV to the list, run the coordinate validator on each FOV generated
6. Additionally, add a cell in the TMA notebook which loops through each centroid of `manual_fov_regions` and runs them through the coordinate validator. Alternatively, we can encapsulate this using a function in `tiling_utils.py` called `verify_manual_fov_coordinates`

**Milestones and timeline**

1. Implementation (TMA + tiled region): 1-2 days
2. Testing (TMA + tiled region): 2-3 days

alex-l-kong:

Closes #197. See the issue for more details.

Again, see issue #197 for more details.


toffy/settings.py
```diff
@@ -9,11 +9,29 @@
 COREG_SAVE_PATH = 'C:\\Users\\Customer.ION\\Documents\\toffy\\toffy\\coreg_params.json'
 FIDUCIAL_POSITIONS = ['top left', 'top right', 'middle left', 'middle right',
                       'bottom left', 'bottom right']
+
 MICRON_TO_STAGE_X_MULTIPLIER = 0.001001
 MICRON_TO_STAGE_X_OFFSET = 0.3116
 MICRON_TO_STAGE_Y_MULTIPLIER = 0.001018
 MICRON_TO_STAGE_Y_OFFSET = 0.6294
 
+STAGE_LEFT_BOUNDARY = -1.03
```
ngreenwald, line 18: How were these values determined? This was what you got on Creed or Drago?

ngreenwald (changes requested):
Looks great. Did you try running all of the notebooks with valid and invalid values to make sure they're caught appropriately?

1. The initial co-registration with bad optical coordinates
2. The initial co-registration with bad stage coordinates
3. The tiled image notebook with FOVs off the slide

toffy/settings.py
alex-l-kong: @ngreenwald yeah I got these from Creed. Should be pretty similar to Drago because the co-registration coefficients on each were about the same.

alex-l-kong:
> Looks great. Did you try running all of the notebooks with valid and invalid values to make sure they're caught appropriately?
> 
> 1. The initial co-registration with bad optical coordinates
> 2. The initial co-registration with bad stage coordinates
> 3. The tiled image notebook with FOVs off the slide

Yup verified that errors are thrown if any coordinate provided is out of bounds.

toffy/settings.py
ngreenwald: Got it. Since the boundaries are checked directly, and don't have any padding added, I think we should be a bit conservative and expand them a bit, similar to the tolerance we add for the other parameters

alex-l-kong: Ok sounds good. A `tol` multiplier of 0.2 is probably not going to work as well for slide boundaries (especially for optical) because the distances between top-bottom and left-right are significantly larger.

We can either use a smaller `tol` multiplier or a fixed `tol` distance (set pretty small, ex. 100 optical pixels or 1 stage coordinate), also open to other ideas.

ngreenwald: Yeah, I think adding/subtracting a constant is a good fix for now. We can revisit if this becomes an issue

ngreenwald (approved):
Nice, looks good

PR #211, opened 2022-08-17
Remove endline in column names from correctly formatted panel

camisowers:

The `convert_panel` function in panel_utils.py reads in files to check if they have a toffy structure. In doing so, the column name 'Stop' becomes 'Stop\n' which interferes with downstream processing such as extracting the bin files. 

Change the column names to their correct previous format before returning. Tested within notebook 3a and fixes the issue.
```
panel.columns = ['Mass', 'Target', 'Start', 'Stop']
return panel
```


ngreenwald (approved):
Sneaky

PR #220, opened 2022-08-20
Add temporary `scripts` directory to track notebook updates

### LINKED ISSUE #78: Automatically generate untracked notebook files upon startup
We currently track the notebooks in git. However, this means that users need to make copies anytime they run those notebooks for their own purposes in order to avoid subsequent merge conflicts

Similar to what the start_docker.sh script does in `ark`, we should add functionality to automatically generate untracked copies of each notebook in a separate folder

alex-l-kong:

Closes #78. This has become more of an issue now that many users have begun running `toffy` notebooks. The following 2 issues have popped up:

1. The main `templates` directory has become crowded, making it difficult to find previous copied notebooks
2. Users make sub-directories inside of `templates`, which led to file path errors

Like `ark-analysis`, we should add a separate `scripts` directory which tracks notebook updates while keeping them encapsulated from the main `templates` directory.

Since we're directly starting Jupyter (and not doing so through Docker like `ark-analysis`), the startup script is called `start_jupyter.sh`, a simplified version of `start_docker.sh` without the additional Docker checks or external drive support.

`update_notebooks.sh` can be copied directly over from `ark-analysis`.

**Remaining issues**

This needs to be tested on Creed to verify the notebooks run smoothly in the new `scripts` directory. Unfortunately, Drago is in use for the foreseeable future, so we likely be unable to set it up there before the cohort finishes.


alex-l-kong:
Just checking to see if everything looks good script-wise, documentation will need to be updated afterwards.

ngreenwald:
I'm a bit on the fence about this one. It definitely solves the git issues, but it adds another step and potential for the user to forget to update and have the wrong version of the notebook. Will you talk to Ben, Hadeesha, and/or Marte and see which they prefer?

alex-l-kong:
@ngreenwald talked with @mkammersgaard and @HPiyadasa, they agree it's a good idea to add this in.

ngreenwald (changes requested):
Will you update the readme as well with the appropriate description of what to do?

ngreenwald (changes requested):
Looks nice. Did you try this for all the configurations?

Making sure it creates the directory when there's nothing there, making sure it doesn't overwrite if -u not supplied, making sure it updates when -u supplied and old version present?

alex-l-kong:
@ngreenwald yup, everything tests out.

alex-l-kong:
@ngreenwald I haven't been able to extensively test this one through the whole `toffy` workflow yet, but because it's also located directly in the `toffy` folder (like `templates`) it shouldn't be an issue.

ngreenwald (approved):

PR #237, opened 2022-09-06
Stitch images function without run file

### LINKED ISSUE #228: Make stitching script independent of run_dir
The current stitch_images function uses the run_json to get image sizes. This is great when once exists, but causes the function to fail if there isn't one. 

We should add behavior to check for a run_json, but if it's not present, instead just get the image size directly by reading a test image

camisowers:

Closes #228.
To allow stitch_images() to work without needing a run file. 

Made the `run_dir` an optional arg in both functions. The get_max_img_size function was altered to check for a run file when a run_dir is provided. If not, it reads in a test file from each FOV folder in the extracted images directory and uses those image sizes to find the maximum.

**Remaining issues**

- [x] edit the notebook documentation to make clear you can have base_dir=None
- [x]  implement for rosetta notebook so there's no need for base_dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,6 @@ The median pulse height (MPH) provides a way to assess the sensitivity of the de
 
 ## Questions?
 
-Feel free to open an [issue](https://github.com/angelolab/toffy/issues) on our GitHub page.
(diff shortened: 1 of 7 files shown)


ngreenwald (approved):
Nice, looks like it simplifies things a bit for the user as well

PR #250, opened 2022-09-08
Notebook 1 normalization checks

### LINKED ISSUE #235: Check inputs for normalization function
The current normalization function in notebook 1 doesn't do any sanity checks of the input before fitting a function

Add in checks before fitting to make sure the inputs are valid.
1. Make sure the directory structure is valid (at least 4 voltages, each in a different folder, with the appropriate bin file in each)
2. Make sure the counts in each FOV for each mass are sufficient (no zero values)

camisowers:

Closes #235.
Add validation checks to the `create_tuning_curve` process. This included checking for a minimum of 4 fovs in the sweep, making sure there is a bin file in each fov folder, and checking that the channel counts for each fov are within the desired range. In addition, added axis labels to the generated tuning curve plot.

* axis labels were added to plot generation but not example curve image 
* raise an error if not enough fovs or missing bin file
* added a `count_range` arg that is used to determine which fovs to exclude, currently excludes values <=0 and >=300,000.
* created two additional test cases
    1. fov with low count that causes the total number to be less than 4 -> fails
    2. fov with low count but still has sufficient number of fovs -> warning

**Remaining issues**

- [x] adjust the count range and add arg to notebook
- [x] discuss more count range implications
- [x] output plot with and without extreme values
- [x] voltage vs max channel count plot

Screen Shot 2022-09-23 at 1 20 08 PM

If the user wants to use the curve with all points, they can set count_range to None and run the cell again. 

Screen Shot 2022-09-22 at 11 17 03 AM


camisowers:
For the voltage vs channel count plot it will show the maximum count value for each FOV, since those are most useful in determining the `count_range` upper threshold.

templates/1_set_up_toffy.ipynb
```diff
@@ -235,7 +235,8 @@
     "# create a new folder with the sweep name\n",
```
ngreenwald, line 235: <p>This is good, but the information about how they are supposed to select the cutoff is buried in the middle (where the counts level off). I would maybe put this as a separate paragraph, or have it come first. Something so that it's more obvious</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/250/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

toffy/normalize.py
```diff
@@ -216,26 +232,37 @@ def combine_run_metrics(run_dir, substring):
...
-
-        # combine tables together
         pulse_heights = pd.read_csv(os.path.join(dir, 'fov-1-scan-1_pulse_heights.csv'))
         channel_counts = pd.read_csv(os.path.join(dir, 'fov-1-scan-1_channel_counts.csv'))
-        combined = pulse_heights.merge(channel_counts, 'outer', on=['fov', 'mass'])
 
+        # check for extreme count values
+        extreme_val = False
+        if count_range:
+            for count in channel_counts.channel_count:
```
ngreenwald, line 255: Rather than a loop over the counts, I would use numpy to compare all the elements at once, and see if any are true

ngreenwald (changes requested):
Looks good. How did you settle on the upper value?

camisowers:
> Looks good. How did you settle on the upper value?

Still needs to be adjusted. Here's a few sweeps from Drago.
This first one is the what we've been using and looks best with no values removed. 
Screen Shot 2022-09-26 at 4 32 43 PM

Here's another one that looks better with some of the lower fov counts removed.
Screen Shot 2022-09-26 at 5 12 33 PM

However, Andrew's curve worked best with the 1.2m cutoff.
I think it's good that we provide the count plot and encourage the user to edit + rerun rather than just using our suggestion.

ngreenwald:
Yeah agreed. Just need to make sure it's very clear from the instructions what they need to do, and how they should be deciding what to remove

ngreenwald (approved):

PR #252, opened 2022-09-08
Rosetta mass 39 naming conflict

### LINKED ISSUE #251: Rosetta ffc correction target conflict

**Describe the bug**
Addresses the bug in #249.
When using rosetta, the `compensate_image_data` function corrects for the channel name `chan_39`. If the user has attached a target to mass 39 (i.e. Potassium), this will cause an error since there will be no `chan_39`, even though the appropriate mass is in fact in the panel.

**Expected behavior**
The `compensate_image_data` function should check for mass 39 in the panel rather than the `chan_39` target.

camisowers:

Closes #251.

Take input `ffc_masses` as an argument and retrieve the specific channel name from the provided `panel_info`.

**Remaining issues**

There's a few functions that similarly have [Noodle](https://github.com/angelolab/toffy/blob/f8f2986d4d943cb023cdc298c50b86d30932ffe1/toffy/rosetta.py#L185) provided as a specified channel name. Is there any instance where there would be a conflict for mass 117, or can we just leave the provided Noodle identifier as is?

### DIFF
--- a/toffy/rosetta.py
+++ b/toffy/rosetta.py
@@ -182,7 +182,7 @@ def get_masses_from_channel_names(names, panel_df):
 def compensate_image_data(raw_data_dir, comp_data_dir, comp_mat_path, panel_info,
                           input_masses=None, output_masses=None, save_format='rescaled',
                           raw_data_sub_folder='', batch_size=1, gaus_rad=1, norm_const=200,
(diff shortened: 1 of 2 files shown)


ngreenwald (approved):
I can't think of a scenario where the user would have something on 117, but if they do put something there I actually think we want to raise an error since their integration range will likely be only a single AMU, where noodle is defined from 117 to 125.

PR #262, opened 2022-09-14
Add _pulse_height.csv file generation to watcher

### LINKED ISSUE #261: Create pulse_heights callback in watcher
Currently, the [normalization notebook](https://github.com/angelolab/toffy/blob/main/templates/4b_normalize_image_data.ipynb) generates pulse_height.csv files for each FOV and then uses them to normalize the images. The process of calculating pulse height values for multiple masses in each bin file can take quite a bit of time (often ~5 mins per FOV).

We would like to create a new FOV level callback for the [real-time monitoring notebook](https://github.com/angelolab/toffy/blob/main/templates/3a_monitor_MIBI_run.ipynb), so that pulse height file generation can be done as soon as the FOV bin and json files are created while the MIBI is running. In addition, the files should be saved to the same directory as the MPH and QC files, rather the the bin file directory where they are currently being saved.

The current logic in the normalization notebook which checks for already existing files before attempting the calculations will remain the same. If the user did not run the real-time processing notebook they can still do image normalization, they will simply spend more time running notebook 4b. 

**Additional context**
There's no need to create an individual notebook for this callback (like the 3b-3e notebooks) because the file generation code already exists in 4b.

camisowers:

Closes #261. Adds pulse height file generation to the real-time monitoring notebook, to shorten the length of time it takes to run the 4b notebook.

Create the FOV level callback `generate_pulse_heights` and implement in notebook 3a. Also add corresponding test cases and validation functions. 

**Remaining issues**

- [x] clean up the test cases and variable names in `fov_watcher_test.py`
- [x] notebook uses mass args `start_offset=0.3` ` stop_offset=0`, but the function `write_mph_per_mass` itself has 0.5 set as defaults for both?

### DIFF
--- a/templates/3a_monitor_MIBI_run.ipynb
+++ b/templates/3a_monitor_MIBI_run.ipynb
@@ -111,10 +111,11 @@
    "source": [
     "fov_callback, run_callback = build_callbacks(\n",
     "    run_callbacks = ['plot_qc_metrics', 'plot_mph_metrics', 'image_stitching'],\n",
-    "    fov_callbacks = ['extract_tiffs'],\n",
+    "    fov_callbacks = ['extract_tiffs', 'generate_pulse_heights'],\n",
     "    tiff_out_dir=extraction_dir,\n",
     "    qc_out_dir=metrics_data_dir,\n",
     "    mph_out_dir=metrics_data_dir,\n",
+    "    pulse_out_dir=metrics_data_dir,\n",
     "    plot_dir=metrics_plot_dir,\n",
     "    save_dir=metrics_plot_dir,\n",
     "    panel=panel,\n",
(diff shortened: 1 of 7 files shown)


templates/3a_monitor_MIBI_run.ipynb
```diff
@@ -111,10 +111,11 @@
    "source": [
```
ngreenwald, line 111: <p>Add a description for what this callback does in the documentation above</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/262/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ngreenwald (changes requested):
Looks good. The reason the start and stop offsets are different is because for normalization, we want a narrow mass range. This ensures the MPH is only calculated from the ions that are actually metal ions, which are in the mass-deficient portion of the peak. 

The function defaults to +/- 0.5 because it was originally written for monitoring the run over time using 97, where everything is organics and it doesn't matter.

ngreenwald (changes requested):
Looks good. Will you try out the actual watcher notebook with some bin files (can be generated already) to make sure everything works

camisowers:
> Looks good. Will you try out the actual watcher notebook with some bin files (can be generated already) to make sure everything works

All good!

ngreenwald (approved):

PR #266, opened 2022-10-05
fix for dimension swap when producing the FoV grid

christianrickert:

To fix an existing bug. - This PR is changing two lines of code that assign the center points to the grid FoVs with X and Y coordinates, respectively.

**How did you implement your changes?**

I have swapped the assignment of the 'fov['centerPointMicrons']['x']' variable with the value for the 'fov['centerPointMicrons']['y'] variable, and vice versa.

https://github.com/angelolab/toffy/blob/7c67a64d80ec282450bd517531dd11d69ade08ea/toffy/tiling_utils.py#L757
```Python3
def generate_tiled_region_fov_list(tiling_params, moly_path: Optional[str] = None):
    """Generate the list of FOVs on the image from the `tiling_params` set for tiled regions
       [...]
    """
[

    # iterate through each region and append created fovs to fov_regions['fovs']
    for region_index, region_info in enumerate(tiling_params['region_params']):
    [...]

        for index, (row_i, col_i) in enumerate(row_col_pairs):
        [...]

            # copy the fov metadata over and add cur_x, cur_y, and name
            fov = copy.deepcopy(tiling_params['fovs'][region_index])
            fov['centerPointMicrons']['x'] = cur_col  # should be 'cur_row'
            fov['centerPointMicrons']['y'] = cur_row  # should be 'cur_col'
```
For details, please see: https://github.com/christianrickert/toffy/commit/273df77db597d889d02e04a388f3fd38af937dbb.

### DIFF
--- a/toffy/tiling_utils.py
+++ b/toffy/tiling_utils.py
@@ -756,8 +756,8 @@ def generate_tiled_region_fov_list(tiling_params, moly_path: Optional[str] = Non
 
             # copy the fov metadata over and add cur_x, cur_y, and name
             fov = copy.deepcopy(tiling_params['fovs'][region_index])
-            fov['centerPointMicrons']['x'] = cur_col
-            fov['centerPointMicrons']['y'] = cur_row
+            fov['centerPointMicrons']['x'] = cur_row
+            fov['centerPointMicrons']['y'] = cur_col
             fov['name'] = fov_names[index]
 
             # append value to fov_regions


alex-l-kong (commented):
@christianrickert were you running into indexing errors when running `toffy` on your cohort?

The reason it was this way is because of the nature of coordinate axes on the MIBI versus in a Python array. The MIBI uses an axis based on standard Euclidean definitions, with the x-axis going left to right and the y-axis going top to bottom (in other words, x indexes columns and y indexes rows). However, in Python (and arrays in general), the axes are flipped, with x indexing rows and y indexing columns.

When generating this run file we need to conform to the MIBI definitions, which is why we assigned `cur_row` to x and `cur_col` to y.

christianrickert:
Looks like some tests failed, expecting swapped coordinate pairs in 'test_generate_tiled_region_fov_list_base'

I don't feel comfortable with changing the tests at this point, since we would make the tests fit the results and not the other way around.

christianrickert:
@alex-l-kong The combination of different coordinate systems is definitely confusing.

I've applied the suggested code change and asked for 2 rows and 7 columns, followed by the deletion of the bottom-right FoVs (on the gold slide). Here's the corresponding post-acquisition SED:

SED_grid

>The MIBI uses an axis based on standard Euclidean definitions, with the x-axis going left to right and the y-axis going top to bottom (in other words, x indexes columns and y indexes rows).

That agrees with my current understanding of the optical coordinate system.

>However, in Python (and arrays in general), the axes are flipped, with x indexing rows and y indexing columns.

I can't follow your logic here, because the coordinates used for the indexing are set with the dictionary keys `['centerPointMicrons']['x']` and `['centerPointMicrons']['y']`.

From what I can see, the problem I addressed (downstream) originated from these earlier lines of code
in the [`generate_tiled_region_fov_list`](https://github.com/angelolab/toffy/blob/7c67a64d80ec282450bd517531dd11d69ade08ea/toffy/tiling_utils.py#L744) function:
```Python
        for index, (row_i, col_i) in enumerate(row_col_pairs):
            # use the fov size to scale to the current row- and col-coordinate
            cur_row = start_row - (row_i * region_info['row_fov_size'])
            cur_col = start_col + (col_i * region_info['col_fov_size'])
```
The `cur_row` decrements its value (top to bottom) to correct for the inversed optical Y-axis coordinates and the `cur_col` increments its value (left to right). That seems logical.
However, variable `cur_row` decrements with the X-coordinate `row_i` (horizontal coordinate) and variable `cur_col` increments with the Y-coordinate `col_i` (vertical coordinate).

I don't see the need to transpose the coordinate system at this point:
Instead, the tuple `(row_i, col_i)` should be `(col_i, row_i)` so that X- and Y-values are assigned to their corresponding column/row indices:

The FoV X-/Y-coordinate pairs are created by the [`generate_x_y_fov_pairs(x_range, y_range)`](https://github.com/angelolab/toffy/blob/7c67a64d80ec282450bd517531dd11d69ade08ea/toffy/tiling_utils.py#L579) function:
```Python
def generate_x_y_fov_pairs(x_range, y_range):
    """Given all x and y coordinates (in microns) a FOV can take,
    generate all possible `(x, y)` pairings
    Args:
        x_range (list):
            Range of x values a FOV can take
        y_range (list):
            Range of y values a FOV can take
    Returns:
        list:
            Every possible `(x, y)` pair for a FOV
    """
    [...]
```
With my current understanding, I would suggest to only fix the tuple variable assignment in the enumeration and to undo the previously suggested change with my pull request.

Please let me know what you think!

alex-l-kong:
@christianrickert agreed, definitely a bit confusing the way the function is written. It doesn't help that MIBI standards require us to specify things in terms of `x` and `y` when `row` and `col` make more sense.

I will have a look into reorganizing `generate_x_y_fov_pairs` sometime early next week, to reduce ambiguity we should rename this function to `generate_row_col_fov_pairs` with the arguments `row_range` and `col_range` instead. Hopefully this clears a few things up.

christianrickert:
Thank you. - Have a great weekend!

ngreenwald:
@christianrickert, did this issue come up when using the notebooks in their current form, or while trying to adapt the helper function for your own purposes?

Basically do we need to update the notebook to make it more clear for regular users, or just the individual function for developers trying to use the underlying functions

PR #269, opened 2022-10-13
Fix extraction logic for run with a single fov crash

### LINKED ISSUE #268: bin_extraction logic fails for run crash during first FOV

**Describe the bug**
We have logic to give a warning and ignore any FOVs with zero sized json files, but when such a FOV is the only one in a run folder ( for example an early crash or a canceled run ), the code throws a generic JSONDecodeError. This is because after removing the bad file from the fov list, an empty list gets passed to `list_moly_fovs`. As a result the function tries to read in all json files in the run and errors out. 

**Expected behavior**
This is a very specific and uncommon case where someone would try to extract from a canceled run, but we should have appropriate logic none the less. 

**To Reproduce**
Screen Shot 2022-10-13 at 2 13 30 PM

camisowers:

Closes #268.

Checks if fov_list is empty before passing to `list_moly_fovs`. 
Also add logic to ignore the run file when checking all json files in the directory for moly indicator. It hasn't caused any issues yet, but technically we don't need to be reading in that large json file.

### DIFF
--- a/toffy/bin_extraction.py
+++ b/toffy/bin_extraction.py
@@ -28,6 +28,10 @@ def extract_missing_fovs(bin_file_dir, extraction_dir, panel,
     if empty_fovs:
         fovs = list(set(fovs).difference(empty_fovs))
 
+    if len(fovs) == 0:
+        warnings.warn(f"No viable bin files were found in {bin_file_dir}", Warning)
+        return
+
     # check for moly fovs
     moly_fovs = list_moly_fovs(bin_file_dir, fovs)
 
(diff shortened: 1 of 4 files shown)


toffy/bin_extraction.py
```diff
@@ -28,6 +28,10 @@ def extract_missing_fovs(bin_file_dir, extraction_dir, panel,
     if empty_fovs:
         fovs = list(set(fovs).difference(empty_fovs))
 
+    if len(fovs) == 0:
+        warnings.warn(f"No viable bin files were found in {bin_file_dir}", Warning)
```
ngreenwald, line 32: Let's add a test to make sure this works as expected. It's very simple, so if we were starting from scratch it wouldn't be worth having a test. However, we should always add a reversion test when fixing bugs to make sure they don't come back.

ngreenwald (changes requested):

ngreenwald (approved):

PR #271, opened 2022-10-18
Make sure loading in convert_panel happens in utf-8

alex-l-kong:

This [line](https://github.com/angelolab/toffy/blob/5732d81255ca164f35add155493416a87e9db2bf/toffy/panel_utils.py#L60) in `panel_utils.py` does not explicitly specify `utf-8` encoding. This can lead to errors when `readlines` is called, such as that in #245.

Specify `encoding='utf-8'` in the `open` call.

### DIFF
--- a/toffy/panel_utils.py
+++ b/toffy/panel_utils.py
@@ -57,7 +57,7 @@ def convert_panel(panel_path):
 
     # read in metadata contained in the first few lines
     col_index = 0
-    with open(panel_path) as r:
+    with open(panel_path, encoding='utf-8') as r:
         lines = r.readlines()
         for index, line in enumerate(lines):
             # line containing the column names we want


toffy/panel_utils.py
```diff
@@ -57,7 +57,7 @@ def convert_panel(panel_path):
 
     # read in metadata contained in the first few lines
     col_index = 0
-    with open(panel_path) as r:
+    with open(panel_path, encoding='utf-8') as r:
```
ngreenwald, line 60: Are there any other places in the repo where this might be a problem? I think most instances are wrapped in the `load_json` function which automatically does it, but worth taking a look to see

ngreenwald (changes requested):

toffy/panel_utils.py
alex-l-kong: This looks like the only remaining place.

ngreenwald (approved):
Sounds good

PR #278, opened 2022-10-26
Add link in notebooks for the ark tiled Image Stitching script

### LINKED ISSUE #276: Link to image stitching code
### Discussed in https://github.com/angelolab/toffy/discussions/275

<div type='discussions-op-text'>

<sup>Originally posted by **akandrew332** October 25, 2022</sup>
I've acquired a bunch of tiled acquisitions.  Is there a script that will stitch these together?</div>

We should add a link in the toffy notebook for tiled image generation to the corresponding code for image stitching which is now in `ark`

camisowers:

Closes #276.
Adds a link to the ark Image Stitching notebook (designed for tiled runs) to both the `2_create_tiled_mibi_run` and `3e_stitch_images` notebooks.

**Remaining issues**

Should we also point them towards section 1 of the `5_rename_and_reorganize` notebook to copy and rename the fov folders?

### DIFF
--- a/templates/2_create_tiled_mibi_run.ipynb
+++ b/templates/2_create_tiled_mibi_run.ipynb
@@ -15,7 +15,7 @@
     "\n",
     "### Background\n",
     "\n",
-    "This notebook allows you to automatically set up large tiled runs of contiguous FOVs. You can specify the dimensions of the tiled image, for example 5x5, and select specific FOVs you wish to exclude from the final tile.\n",
+    "This notebook allows you to automatically set up large tiled runs of contiguous FOVs. You can specify the dimensions of the tiled image, for example 5x5, and select specific FOVs you wish to exclude from the final tile. To stitch these FOV images back into their tiled shape later, use the [Image Stitching notebook](https://github.com/angelolab/ark-analysis/blob/main/templates/Image_Stitching.ipynb) in the ark-analysis repo.\n",
     "\n",
     "The script expects that you have already generated and moved the necessary files into the appropriate directory before starting.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


ngreenwald:
There are going to be many people who want to generate stitched images of the raw data as soon as it comes off the instrument. How much work would it be to make a toffy-specific version of the stitching script that takes in the run JSON file, and the original fov-1, fov-2, etc names, and generates a correct stitched image? I'm thinking it could make a dictionary or something mapping the fov-1 names to the 'true' names, and then use that dictionary combined with the existing function in `ark` to use the non-renamed FOVs

camisowers:
TLDR: the ark functions would require adjustments no matter what, but I think it would be easiest to make the load function compatible with toffy and then adjust [toffy.image_stitching.stitch_images](https://github.com/angelolab/toffy/blob/3fbe93478ded3ccedafd6b650c2a489020319459/toffy/image_stitching.py#L60) to have a default `tiled=False` arg that we can adjust the image loading logic for. 

Using the ark function `stitch_images_by_shape` would require quite a bit of altering since it does various directory/file validation checks and has a lot of extra logic for the segmentation/clustering dir structures. We wouldn't need most of that since the fov names would be provided from the json file and can be checked while generating a fov names dict.

The `load_tiled_img_data` function would only require two small tweaks:
1. allowing fov_list to take either a list or dictionary and subsequent logic to differentiate between RnCm name and fov-1 name needed for loading in files
2. adding an arg to specify expected fov names and moving `expected_fovs = get_tiled_fov_names(fov_list)` out to the `stitch_images_by_shape` function instead. 

Once these are done, the load function could be called with the dictionary of fov-1 / RnCm names, and we can just write a toffy equivalent to the stitching function that does any name validation and dict creation there.

ngreenwald (approved):
Got it, sounds good

PR #285, opened 2022-11-07
Tiled stitching for toffy

### LINKED ISSUE #279: Create toffy-native functionality for stitching images
Given that many users will want to look at their tiled images immediately after they are generated, rather than waiting until after rosetta is run, we should have the option to generate stitched images directly in toffy. 

See below for the proposed plan:

TLDR: the ark functions would require adjustments no matter what, but I think it would be easiest to make the load function compatible with toffy and then adjust [toffy.image_stitching.stitch_images](https://github.com/angelolab/toffy/blob/3fbe93478ded3ccedafd6b650c2a489020319459/toffy/image_stitching.py#L60) to have a default `tiled=False` arg that we can adjust the image loading logic for. 

Using the ark function `stitch_images_by_shape` would require quite a bit of altering since it does various directory/file validation checks and has a lot of extra logic for the segmentation/clustering dir structures. We wouldn't need most of that since the fov names would be provided from the json file and can be checked while generating a fov names dict.

The `load_tiled_img_data` function would only require two small tweaks:
1. allowing fov_list to take either a list or dictionary and subsequent logic to differentiate between RnCm name and fov-1 name needed for loading in files
2. adding an arg to specify expected fov names and moving `expected_fovs = get_tiled_fov_names(fov_list)` out to the `stitch_images_by_shape` function instead. 

Once these are done, the load function could be called with the dictionary of fov-1 / RnCm names, and we can just write a toffy equivalent to the stitching function that does any name validation and dict creation there.

_Originally posted by @camisowers in https://github.com/angelolab/toffy/issues/278#issuecomment-1292861924_

camisowers:

Closes #279. Gives users the option to stitch images by original tiled shape rather than just acquisition order. 

Detailed in https://github.com/angelolab/ark-analysis/issues/795. Adds a helper function to get the RnCm names from the run file and store them in a dictionary along with their corresponding default fov name. Adjusts `image_stitching.stitch_images()` to have a tiled bool arg and updated the notebook accordingly.

**Remaining issues**
There's a logic issue if a fov has more than one scan, but for tiled runs I don't think that should be an issue (?). Right now it only loads in the first scan for each fov.

### DIFF
--- a/templates/3e_stitch_images.ipynb
+++ b/templates/3e_stitch_images.ipynb
@@ -8,7 +8,7 @@
     "# Stitch Images Notebook\n",
     "## This notebook is an example: create a copy before running it or you will get merge conflicts!\n",
     "\n",
-    "This notebook will create a single stitched image for each channel in your panel across all of the FOVs in your run, allowing you to quickly inspect your data. Images will be stitched together based on acquisition order. If you would like to generate images based on the original tiled shape specified in [2_create_tiled_mibi_run](./2_create_tiled_mibi_run.ipynb), use the [Image Stitching notebook](https://github.com/angelolab/ark-analysis/blob/main/templates/Image_Stitching.ipynb) in the ark-analysis repo instead.\n",
+    "This notebook will create a single stitched image for each channel in your panel across all of the FOVs in your run, allowing you to quickly inspect your data. By default, images will be stitched together based on acquisition order. If you would like to generate images based on the original tiled shape specified in [2_create_tiled_mibi_run](./2_create_tiled_mibi_run.ipynb), make sure to specify so with `tiled_image=True`.\n",
     "\n",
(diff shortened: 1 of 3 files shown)


ngreenwald (changes requested):
Looks good, did you run the notebook on some real data to make sure it works?
We should check:
1) tiled=False for a tiled image
2) tiled = True for a tiled image
3) tiled = False for a non-tiled image (the old functionality)
3) Tiled = True for a non-tiled image (does this raise an error?)

camisowers:
Notebook checks:
- [x] tiled=False for a tiled image -> will just stitch in acquisition order
- [x] tiled = True for a tiled image
- [x] tiled = False for a non-tiled image
- [x] Tiled = True for a non-tiled image -> **(error caught from `load_utils.get_tiled_fov_names()` and custom error added)**

Function also now excludes any moly fovs from the names dictionary.

ngreenwald (approved):
Looks good!

PR #293, opened 2022-11-17
Make sure Rosetta converts images to `float` prior to running

### LINKED ISSUE #292: Load images in as `float32` prior to compensation
**Describe the bug**

Because data is now loaded in using the default type of the images in the directory, we can no longer explicitly cast the type to `flaot32`. This causes issues with Rosetta compensation, as the input images may be `uint32`, which causes `numpy` operation type errors. Functions like `gaussian_filter` propagate this error rather than correct for this.

After making a call to `load_imgs_from_tree`, we should explicitly add a type cast to `np.float32`.

**Expected behavior**

The pipeline should run seamlessly.

**To Reproduce**

Use a cohort of `uint32` images on Rosetta.

alex-l-kong:

Addresses #288 and closes #292. Because the functionality of `load_imgs_from_tree` was updated to not use explicit `dtype` loading, cohorts of `uint32` will not work due to floating point incompatibilities  caused by channel coefficient subtraction. This `dtype` gets propagated through functions such as `gaussian_filter` rather than being corrected.

We'll need to add an explicit correction for this loading prior to running them through Rosetta

Since we can't tell `load_imgs_from_tree` which `dtype` to use, we'll instead add an explicit `.astype(np.float32)` cast right after the call.

### DIFF
--- a/toffy/rosetta.py
+++ b/toffy/rosetta.py
@@ -247,6 +247,9 @@ def compensate_image_data(raw_data_dir, comp_data_dir, comp_mat_path, panel_info
                                                     channels=acquired_targets,
                                                     img_sub_folder=raw_data_sub_folder)
 
+        # convert to float32 for gaussian_filter and rosetta compatibility
+        batch_data = batch_data.astype(np.float32)
+
         # blur data
         if gaus_rad > 0:
             for j in range(batch_data.shape[0]):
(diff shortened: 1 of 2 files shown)


alex-l-kong:
@ngreenwald adjusted the tests so that it now runs `compensate_image_data` using `np.uint32` images. Confirms it passes with the explicit cast after `load_imgs_from_tree` and fails without it.

PR #300, opened 2022-12-01
Update Rosetta test directory setup

### LINKED ISSUE #291: Rework Rosetta directory structure to support subsequent rounds of compensation
**Relevant background**

Oftentimes, one round of Rosetta will not be enough to remove all unwanted artifacts from an image. Sometimes an additional round of Rosetta will be needed to smooth out specific channels.

The first step is to rework the folder structure of Rosetta to prevent them from blowing up in size. This will also make it easier to navigate the folder to find the relevant files needed.

**Design overview**

The new folder structure will be as follows:

Screen Shot 2022-11-15 at 12 01 09 PM

There are a few motivations for this:

- The new Rosetta naming scheme will make it easier to track which matrix was used for which input and output channel(s) are being used. This is important because we'll need to provide users and easy way to distinguish between the v1 and v2 (and potentially multiple v2) matrices.
- Removing the `compensated_data_{mult}` and `stitched_images` folders saves a ton of space. This will be extremely helpful because Creed and Drago reach full capacity otherwise.

**Code mockup**

`create_rosetta_matrices` will need to take in `current_channel_name` and `output_channel_names` (via `generate_rosetta_test_imgs`) to generate the nomenclature needed. The new signature will be:

```
create_rosetta_matrices(default_matrix, save_dir, multipliers, current_channel_name='Noodle', output_channel_name=None, masses=None):
    """Creates a series of compensation matrices for evaluating coefficients
    Args:
        default_matrix (str): path to the rosetta matrix to use as the default
        save_dir (str): output directory
        multipliers (list): the range of values to multiply the default matrix by
            to get new coefficients
        current_channel_name (str):
            channel being adjusted, default Noodle
        output_channel_names (list):
            subset of the channels to compensate for, default None is all
        masses (list | None): an optional list of masses to include in the multiplication. If
            only a subset of masses are specified, other masses will retain their values
            in all iterations. If None, all masses are included
    """
```

The files will now be saved using: 

```
# in the case output_channel is None, it will be set to `all`
mult_matrix.to_csv(os.path.join(save_dir, current_channel_name + '_' + '_'.join(output_channel) + '_' + + '_mult_%s.csv' % (str(i))))
```

Additionally, we'll need to create a helper function to remove the `compensated_image_data` and `stitched_images` folders. We'll create add a function at the end of `generate_rosetta_test_imgs` called `clean_rosetta_dir` with signature:

```
def clean_rosetta_dir(cohort_path: str):
    """Removes extraneous `compensated_image_data` and `stitched_images` folders

    Args:
        cohort_path (str):
            Path to rosetta cohort
    """
```

This function should be called at the end of `compensate_image_data`. However, since this function gets called twice (once for verification, we should add a flag to `compensate_image_data` called `remove_comp_folders` which is by default set to `False`, but we overwrite to `True` on the final call.

**Required inputs**

A summary of the new parameters described above:

- Updated `create_rosetta_matrices`: `current_channel_name` + `output_channel_names`
- Updated `compensate_image_data`: `remove_comp_folders`

No additional files needed.

**Output files**

The nomenclature of the new rosetta matrices is `{src}_{out}_rosetta_matrix_mult_{mult}.csv`.

`compensated_data_{mult}` and `stitched_images` folders will be removed.

**Timeline**
- [ ] A couple days
- [X] A week
- [ ] Multiple weeks. For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

11/19/22

11/22/22

alex-l-kong:

Closes #291.

See issue for details.

One minor change is that we don't actually want to put the cleanup process directly in `compensate_image_data`, since we should only run this cleanup for Rosetta testing. We instead call the cleanup function separately after `add_source_channel_to_tiled_image` is added.

### DIFF
--- a/templates/4a_compensate_image_data.ipynb
+++ b/templates/4a_compensate_image_data.ipynb
@@ -233,16 +233,8 @@
     "output_dir = os.path.join(rosetta_testing_dir, cohort_name, folder_name + '-stitched_with_' + current_channel_name)\n",
     "os.makedirs(output_dir)\n",
     "rosetta.add_source_channel_to_tiled_image(raw_img_dir=img_out_dir, tiled_img_dir=stitched_dir,\n",
-    "                                          output_dir=output_dir, source_channel=current_channel_name, max_img_size=img_size)"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "e4f946ee-ee8c-4d05-b0f5-4d7390914bcc",
-   "metadata": {},
-   "outputs": [],
-   "source": [
+    "                                          output_dir=output_dir, source_channel=current_channel_name, max_img_size=img_size)\n",
+    "\n",
     "# remove the intermediate compensated_data_{mult} and stitched_image dirs\n",
     "rosetta.clean_rosetta_test_dir(folder_path)"
    ]
--- a/templates/4a_compensate_image_data.ipynb
+++ b/templates/4a_compensate_image_data.ipynb
@@ -237,6 +237,17 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e4f946ee-ee8c-4d05-b0f5-4d7390914bcc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# remove the intermediate compensated_data_{mult} and stitched_image dirs\n",
+    "rosetta.clean_rosetta_test_dir(folder_path)"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "id": "21d6a3fe-19a1-4df1-a952-6c4b2bb7dc07",
    "metadata": {},
@@ -400,7 +411,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "toffy38",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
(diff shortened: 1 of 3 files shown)


templates/4a_compensate_image_data.ipynb
```diff
@@ -233,7 +233,10 @@
     "output_dir = os.path.join(rosetta_testing_dir, cohort_name, folder_name + '-stitched_with_' + current_channel_name)\n",
```
HPiyadasa, line 233: <p>Are the different rosetta matrices generated with the multipliers saved or deleted? </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

alex-l-kong: <p>They're saved. Running <code>os.listdir(folder_path)</code> after this cell yields:</p><pre class="ql-syntax" spellcheck="false">['Noodle_all_commercial_rosetta_matrix_mult_1.csv',
 'Noodle_all_commercial_rosetta_matrix_mult_2.csv',
 'Noodle_all_commercial_rosetta_matrix_mult_0.5.csv'] 
</pre>

HPiyadasa (approved):

ngreenwald (changes requested):
@HPiyadasa Did you try running this version of the code to make sure the formatting of the .csv names fits with what you want, that the right folders are deleted, etc?

alex-l-kong:
> @ngreenwald Tested the code and works well. Will there be additional PR opened to address other aspects from the original design doc?

Yes there will be, part 2 will address the underlying V2 implementation.

ngreenwald (approved):
Looks good! Can get started on the design doc for part 2 when you have a chance.

PR #313, opened 2023-01-31
Add ability to override run callbacks to process per FOV

### LINKED ISSUE #306: 3a Monitor run notebook does not plot until the end of run
**Describe the bug**
Watcher script finds the bin files and extracts the tiffs and generates the MPH metrics (found in appropriate dir) but does not **plot** graphs until the run is finished

**Expected behavior**
Plots should be generated (channel stats, MPH plot etc) and updated every time a new FOV is found.

### LINKED ISSUE #311: Add option to execute certain run callbacks per FOV
**Relevant background**

Addresses issue #306. For very large cohorts, it isn't convenient to wait until the end of a run for certain run callbacks to execute. This is especially relevant for plotting functions such as `plot_qc_metrics` and `plot_mph_metrics`. If a long run is failing, the user should be told ASAP so they can stop and make the necessary adjustments.

**Design overview**

These changes concern the watcher script in notebook 3a. This function should now contain an option to pass certain run callback names that the user can use to "override" a run callback as a FOV callback (in quotes because we won't explicitly be changing it to a FOV callback). We will call these "intermediate callbacks".

This means internal changes to the way callbacks are handled. We will need to create a separate callback function for these intermediate callbacks and ensure that these get run after the end of every FOV (they cannot be handled as normal FOV callbacks because these intermediate callbacks still depend on certain FOV callbacks completed).

Since the potential run callbacks that will need to run per FOV are all plotting functions, the plotting functionality should also be updated to reflect a per-FOV update scheme.

**Code mockup**

`build_callbacks`

This will now take an argument `intermediate_callbacks`, which the user can pass the plotting functions (and other potentially) to. Should the user not specify any, it defaults to `None`.

The loop to check for prerequisites will need to be updated:

```python
callbacks_with_prereq = run_callbacks + intermediate_callbacks if intermediate_callbacks else run_callbacks[:]
for run_cb in callbacks_with_prereq:
```

Due to `run_callbacks` wrapping all the run callbacks in loop format inaccessible on an individual level, we'll need to define a separate variable `intermediate_callbacks` that essentially functions like `run_callbacks`:

```python
intermediate_callback = None
if intermediate_callbacks:
    def intermediate_callback(run_folder: str):
        callback_obj = RunCallbacks(run_folder)

        for run_cb in intermediate_callbacks:
            if cb := getattr(callback_obj, run_cb, None):
                cb(**kwargs)
            else:
                # unreachable...
                raise ValueError(
                    f'Could not locate attribute {run_cb} in RunCallbacks object'
                )

return fov_callback, intermediate_callback, run_callback
```

`fov_watcher.py`

`start_watcher` will need to take this `intermediate_callback` function, which then passes it to `FOV_EventHandler` and assigns it to `self.inter_func`.

After `fov_func` runs in `on_created`, we should then run `self.inter_func(self.run_folder, point_name)`. This should be fine because all the prerequisite `fov_func` functions should have completed for that `point`. Alternatively, we could do this in `check_complete` to maintain the definition of `inter_func` as essentially a run function that gets overridden as a FOV function.

_Visualization_

One robust method is, for each `intermediate_callback` function call, we append new subplots the main figure. [`matplotlib.gridspec`](https://matplotlib.org/stable/api/_as_gen/matplotlib.gridspec.GridSpec.html) allows us to do this dynamically. This would require modifications to `visualize_mph`, and `visualize_qc_metrics`. We can use `visualize_metric_create_plot` as a wrapper function which delegates the plotting functionality to some utility functions, thus making it work for 1,2, or even 40 subplots.

```python

fig = plt.figure(constrained_layout=True, figsize=fig_size)

gs = gridspec.GridSpec(nrows=subplot_nrows, ncols=subplot_ncols, figure = fig)

for ax_idx, data for enumerate(combined_data):
    
    ax = fig.add_subplot(gs[ax_idx])

    visualize_metric_create_plot(ax = ax)


plt.show() # refreshes the plot for each intermediate_callback
```

```python
def visualize_metric_create_plot(metric: Optional["qc", "mph"], ax: plt.Axes, out_dir: pathlib.Path, **kwargs) -> None:
    # perform any necessary processing here that's shared between "qc" and "mph"

    if metric == "qc":
        visualize_qc_metrics(..., ax = ax) # Specify the axis so the actual plot generation with seaborn / matplotlib knows where to add it to
    if metric == "mph":
        visualize_mph(..., ax = ax)
    
```

**Required inputs**

The user will need to specify `intermediate_callbacks` as a parameter to `build_callbacks`, which will then also need to be passed to `start_watcher`.

**Output files**

Same as before.

**Timeline**

- [ ] A couple days
- [ ] A week
- [X] Multiple weeks. For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

TBD

TBD

alex-l-kong:

Closes #311. Closes #306. Closes #311 

See design doc in #311.


alex-l-kong:
@srivarra all up and running now! The failing tests are from the old MIBItracker API code, we don't have to worry about that for this PR.

ngreenwald:
@alex-l-kong and @srivarra, will you guys take a look through the whole thing to confirm you guys are both happy with the final result?

toffy/mph_comp.py
```diff
@@ -152,7 +152,7 @@ def visualize_mph(mph_df, out_dir, regression: bool = False):
         io_utils.validate_paths(out_dir)
 
     # visualize the median pulse heights
-    plt.style.use('dark_background')
+    # plt.style.use('dark_background')
     # plt.title('FOV total counts vs median pulse height')
```
alex-l-kong, line 156: These comments will need to be removed.

srivarra: @alex-l-kong Should be good now, just pushed the changes.

alex-l-kong:
@srivarra and I reviewed, looks good on our end.

ngreenwald (approved):

toffy/qc_comp.py
```diff
@@ -264,14 +266,11 @@ def sort_bin_file_fovs(fovs, suffix_ignore=None):
     )
 
 
-def compute_qc_metrics(bin_file_path, extracted_imgs_path, fov_name,
+def compute_qc_metrics(extracted_imgs_path, fov_name,
```
camisowers, line 269: Note that the [generate_qc](https://github.com/angelolab/toffy/blob/d088f740952683fcf8c84367cc322adcd0422c92/toffy/watcher_callbacks.py#L171) FOV callback actually calls `compute_qc_metric_direct` itself, not this function. I think this was leftover from when the function above used the bin files for calculation, so Adam made a "direct" function to skip that re-extraction while running watcher. We should probably adjust the dosctrings of both functions (and possibly the helper function name) to make clear what's going on.

camisowers (commented):
Looks good! Just one minor comment about the logic of `compute_qc_metrics_direct`.

PR #326, opened 2023-03-15
Ensure normalization doesn't happen for all-zero pulse height channels

### LINKED ISSUE #307: 4b, Normalization script sometimes produces an error when smoothing outliers

**Describe the bug**
The error is new_value is NaN in normalize.py around line 487.  It happens when both previous_vals and subsequent_vals are empty (Not certain how this happens, could be a very sparse file with little signal). Our temporary solution is to assign an intensity value of 0 to the offending outlier.

**Expected behavior**
There needs to be a check for whether both previous and subsequent arrays are empty.

**To Reproduce**
Please either copy/paste or screenshot all of the code you ran which produced the error. In addition, copy the full error message. 

The error happens on 
line 490: smoothed_vals[outlier] = new_val 
Because the array expects a value, it does not accept a NaN.

This is a temporary solution to the problem:
Screen Shot 2023-01-10 at 10 30 57 AM

alex-l-kong:

Closes #307. When the pulse heights for a particular mass are all zero, normalization will fail, specifically during the outlier calculations. We'll need to be careful about how we handle all-zero masses further down the pipeline as well, because we need to account for divisions by zero.

Four changes were made in `normalize.py`:

1. `create_fitted_pulse_heights_file`: if we detect that all `mph_vals` for a particular `mass` is zero, we display a warning to the user and skip `fit_mass_mph_curve`. This also means we won't have a `_norm_func.json` file for this particular mass, which leads to...
2. `create_fitted_mass_mph_vals`: if a particular mass doesn't have a corresponding `_norm_func.json` file, we set `pulse_height_df.loc[mass_idx, 'pulse_height_fit']` to `0.` and move on.
3. `normalize_image_data`: prior to passing `pulse_height_fov['pulse_height_fit'].values` into `norm_func`, we have to suppress warnings with the following:

```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
```

because passing a `pulse_height_fit` of `0.` into a `log` function will throw a `numpy` warning that we don't want to get displayed several times to the user. Later, we do correct for any `np.inf, `-np.inf`, and `np.nan` values that get spit out by `norm_func`, defaulting them to 0.

4. `normalize_fov`: to prevent divisions from 0 causing `np.inf`, we explicitly use the following:

```
normalized_images = np.divide(
    img_data, norm_vals.reshape((1, 1, 1, len(norm_vals))), out=np.zeros_like(img_data, dtype=img_data.dtype), where=norm_div != 0
)
```

**Remaining issues**

We'll need to get @dtebaykin to test this out, since he was the one who originally opened this issue.


alex-l-kong:
@ngreenwald as mentioned, I'll be asking Dmitry to test out this PR as well.

ngreenwald (changes requested):
Sure thing, once everyone else is happy you can re-request

toffy/normalize.py
```diff
@@ -565,8 +565,15 @@ def create_fitted_mass_mph_vals(pulse_height_df, obj_func_dir):
...
 
     for mass in masses:
-        # load channel-specific prediction function
+        # if channel-specific prediction function does not exist, set to 0
         mass_path = os.path.join(obj_func_dir, str(mass) + '_norm_func.json')
+        mass_idx = pulse_height_df['mass'] == mass
+
+        if not os.path.exists(mass_path):
+            pulse_height_df.loc[mass_idx, 'pulse_height_fit'] = 0.
+            continue
```
srivarra, line 574: Does the `continue` here do something different than if it was not there?

toffy/normalize.py
```diff
@@ -612,6 +618,12 @@ def create_fitted_pulse_heights_file(pulse_height_dir, panel_info, norm_dir, mas
     # loop over each mass, and fit a curve for MPH over the course of the run
     for mass in masses:
         mph_vals = pulse_height_df.loc[pulse_height_df['mass'] == mass, 'pulse_height'].values
+
+        # only create a _norm_func file if the mph_vals are non-zero
+        if np.all(mph_vals == 0):
+            warnings.warn("Skipping normalization for mass %s with all zero pulse heights" % mass)
+            continue
```
srivarra, line 625: Same `continue` question here.

toffy/normalize.py
```diff
@@ -641,9 +653,15 @@ def normalize_fov(img_data, norm_vals, norm_dir, fov, channels, extreme_vals):
...
+    # correct images and save, ensure that no division by zero happens
     norm_vals = norm_vals.astype(img_data.dtype)
-    normalized_images = img_data / norm_vals.reshape((1, 1, 1, len(norm_vals)))
+    norm_div = norm_vals.reshape((1, 1, 1, len(norm_vals)))
+    normalized_images = np.divide(
+        img_data,
+        norm_div,
+        out=np.zeros_like(img_data, dtype=img_data.dtype),
+        where=norm_div != 0
+    )
```
srivarra, line 664: One alternative method that could be used here is [`masked arrays`](https://numpy.org/doc/stable/reference/maskedarray.html). With this we can create a mask over the `norm_vals` array which will filter values of interest. In this case, it would be `0`.

For example:

```python
import numpy as np
norm_vals_masked = np.ma.masked_values(
	x = norm_vals, # array to mask
	value = 0, # the value to mask,
)
```

This would effectively block `0` values from getting passed into`norm_func` on lines 727.

srivarra (changes requested):
Looks good, just a question about `continue` and one suggestion about handling the troublesome `0` values for division / logarithm.

toffy/normalize.py
```diff
@@ -701,9 +719,18 @@ def normalize_image_data(img_dir, norm_dir, pulse_height_dir, panel_info,
...
+        # suppress warnings that can be caused by passing 0 or inf values into norm_func
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            norm_vals = norm_func(pulse_height_fov['pulse_height_fit'].values)
+
+        # nan or inf norm_vals can appear by passing zeros to norm_func, force back to 0
+        norm_vals[norm_vals == np.inf] = 0
+        norm_vals[norm_vals == -np.inf] = 0
+        norm_vals[np.isnan(norm_vals)] = 0
+
```
camisowers, line 733: Kinda roundabout to suppress warnings and correct the invalid results after the calculation. Maybe consider changing `def log()` and `def exp()` in [create_objective_function()](https://github.com/angelolab/toffy/blob/e6549638c933c19d1c19adfeb3c4783f18e7a348/toffy/normalize.py#L104) to always return a 0 when x<=0. Then it won't try to use np.log incorrectly and no warnings raised.

camisowers (commented):
Code looks clean to me! Just one suggestion for a different place to handle the out of domain values for `np.log()`.

toffy/normalize.py
alex-l-kong: See above.

toffy/normalize.py
alex-l-kong: Sounds good! Was a bit hesitant playing around with the actual objective functions at first, but this should work!

alex-l-kong:
Dmitry has tested this out and given the green light.

srivarra (approved):
Looks good!

camisowers (approved):

src/toffy/normalize.py
```diff
@@ -707,9 +728,11 @@ def normalize_fov(img_data, norm_vals, norm_dir, fov, channels, extreme_vals):
             "recommended: {}".format(fov, bad_channels)
         )
 
-    # correct images and save
+    # correct images and save, ensure that no division by zero happens
     norm_vals = norm_vals.astype(img_data.dtype)
-    normalized_images = img_data / norm_vals.reshape((1, 1, 1, len(norm_vals)))
+    norm_div = norm_vals.reshape((1, 1, 1, len(norm_vals)))
+    norm_vals_masked = np.ma.masked_values(x=norm_vals, value=0)
+    normalized_images = np.ma.divide(img_data.values, norm_div).filled(0)
```
ngreenwald, line 735: Is this divide by zero stuff happening as a result of skipping the masses with 0 MPH, or it was already a problem but it's only now becoming apparent?

ngreenwald (changes requested):

src/toffy/normalize.py
alex-l-kong: This is happening because we're skipping masses with 0 MPH.

ngreenwald (approved):

PR #342, opened 2023-04-06
Mass proficient extraction for NAS storage

### LINKED ISSUE #322: NAS-friendly extraction
The current practice of storing all the bin files is likely not necessary for the long term. In order to ensure everyone is confident that they can remove their bin files after a year, we'll want to have extracted all potentially relevant information from all bin files before we delete them. 

We'll want to extract the mass-proficient (x+0.3) in addition to our current deficient (x-0.3) setup. In addition, we'll want to extract all potentially relevant antibody/naturally abundant channels. We'll then need to decide how we handle the proficient and deficient copies of each channel so that rosetta and normalization work as expected.

### LINKED ISSUE #327: Extract proficient data to account for NAS bin file deletion
**Relevant background**

The amount of `.bin` files on the NAS is taking up too much space (at >80% capacity, with `.bin` files accounting for at least 60% of that), so we'll need to ensure their removal after a year. However, the information from these `.bin` files should still be maintained on the NAS. One additional important piece of information that will need to be extracted is mass proficient data.

**Design overview**

The main changes to this pipeline concern processes run by both notebooks `3a` (for those running the watcher) and `3b` (for those extracting after a run). This is because both processes call the bin file extraction function.

Currently, the default is to only extract mass-deficient data (-0.3, 0). To extract mass-proficient data, we'll need to run a separate extraction from (0, 0.3). Note that it is not possible to accomplish this by extracting data from (-0.3, 0.3).

With regards the panel files, which contain these integration ranges, these are by default also in mass-deficient ranges. For mass-proficient extraction, we'll need to ensure that these ranges get adjusted accordingly.

**Code mockup**

- _`watcher_callbacks.FovCallbacks` attributes_

A new attribute, `__fov_data_prof: xr.DataArray = field(default=None, init=False)` will need to be added. It is best to maintain a separate variable for the extracted positive data as opposed to just overwriting `__fov_data` itself.

- _`panel_utils.generate_prof_panel`_

This will be a new hidden function to assist with defining the panel for mass-proficient extraction, implemented as such:

```
def generate_prof_panel(panel: pd.DataFrame):
    panel_positive = panel.copy()

    # extract only rows where 'Start' - 'End' = -0.3
    panel_positive_rows = panel_positive[panel_positive['Start'] - panel_positive['End'] == -0.3].index.values

    # explicitly modify the integration range to (0, 0.3)
    panel_positive['Start'] = panel_positive.loc[panel_positive_rows, 'Stop'].copy()
    panel_positive['Stop'] = panel_positive.loc[panel_positive_rows, 'Start'].copy() + 0.3

    return panel_positive
```

- _`watcher_callbacks.FovCallbacks._generate_fov_data`_

This function will need to make a second call to `extract_bin_files` to assign the mass-proficient data to `__fov_data_prof`.  The only changes to make for this call is to set `panel=panel_utils.generate_prof_panel(panel)` and to assign it to `self.__fov_data_prof`.

Note that `panel_utils` is not currently imported in `watcher_callbacks.py`, so this will also need to be added.

- _`watcher_callbacks.FovCallbacks.extract_tiffs`_

The check for `self.__fov_data is None` should also check `self.__fov_data_prof is None` for safety purposes.

This function will need to make a second call to `_write_out` to save the data in `__fov_data_prof`. The only change to make for this call is to set `out_dir=tiff_out_dir + '_proficient'`. This is needed because the data should be written to a separate proficient folder prefixed by the same run name.

- _`watcher_callbacks.generate_pulse_heights`_

We'll need to extract pulse heights for proficient data as well. This will require an additional call to `write_mph_per_mass` inside `generate_pulse_heights`. The call we'll make is:

```
write_mph_per_mass(
    base_dir=self.run_folder,
    output_dir=pulse_out_dir,
    fov=self.point_name,
    masses=panel["Mass"].values,
    start_offset=kwargs.get("mass_start_prof", 0),
    stop_offset=kwargs.get("mass_stop_prof", 0.3),
    proficient=True,
)
```

to ensure the range goes from 0 to 0.3.

Note that folder and file handling is done differently for this function, so a few additional tweaks need to be made inside `normalize.write_mph_per_mass`.

- _`normalize.write_mph_per_mass`_

This function should take another argument called `proficient` of type `bool`. This is simply needed to control the name of the file that gets saved:

```
pulse_heights_file = (
    fov + "_pulse_heights_proficient.csv" if proficient else fov + "_pulse_heights.csv"
)
```

This adds separation between mass deficient and proficient data files.

- _`3b_extract_images_from_bin.ipynb`_

**NOTE: this replaces the previous method where the mass proficient data was extracted in a separate call inside `bin_extraction.extract_missing_fovs`. Needed to change to ensure an easier scheme to account for potentially different missing FOVs inside the deficient and proficient extraction folders.**

An additional call to `extract_missing_fovs` will need to be made on the mass proficient folder. Simply copy the call for the default (mass-deficient) data and add a few changes:

```
extract_missing_fovs(base_dir, extraction_dir + '_proficient', generate_prof_panel(panel))
```
`generate_prof_panel` isn't currently imported by this notebook, so the import statement should be added.

_`4b_normalize_image_data.ipynb`_

If the pulse heights haven't already been extracted by notebook 3a, they get handled in this notebook.

The control flow logic should be modified to 

```
# get all FOVs
fovs = list_folders(os.path.join(rosetta_base_dir, run_name), substrs='fov-')

# loop over each FOV
for fov in fovs:
    # generate mph values
    mph_file_path = os.path.join(mph_run_dir, fov + '_pulse_heights.csv')
    mph_file_path_prof = os.path.join(mph_run_dir, fov + '_pulse_heights_proficient.csv')
    
    # base MPH normalization
    if not os.path.exists(mph_file_path):
        normalize.write_mph_per_mass(base_dir=os.path.join(bin_base_dir, run_name), output_dir=mph_run_dir, 
                                     fov=fov, masses=panel['Mass'].values, start_offset=0.3, stop_offset=0)

    # mass proficient MPH normalization (for long-term storage)
    if not os.path.exists(mph_file_path_prof):
        normalize.write_mph_per_mass(base_dir=os.path.join(bin_base_dir, run_name), output_dir=mph_run_dir, 
                                     fov=fov, masses=panel['Mass'].values, start_offset=0, stop_offset=0.3,
                                     proficient=True)
```

We add another check if the proficient pulse heights file exists or not, and if the latter, create it using the new `start_offset` and `stop_offset`.

**Required inputs**

Same inputs as before.

**Output files**

A new folder, `tiff_out_dir + '_proficient'`, will be created with the mass-proficient images extracted.

**Timeline**

Once I get back from my Europe trip, it should take:

- [ ] A couple days
- [ ] A week
- [X] Multiple weeks. For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

4/17 (in Europe from 3/21-4/2)

4/20 (lol)

alex-l-kong:

Closes #327. Closes #322.

See issue link.

**Remaining issues**

Need to test this on Christine's run next week.


srivarra (approved):
LGTM!

camisowers:
Agreed, looks good!

src/toffy/watcher_callbacks.py
```diff
@@ -141,11 +146,26 @@ def _generate_fov_data(
             time_res=time_res,
         )
 
+        panel_prof = generate_prof_panel(panel)
```
ngreenwald, line 149: I think it would be worth having a flag that controls whether this happens or not.

ngreenwald (changes requested):
Looks good. You should also confirm that the offsets are being handled correctly for mass extraction, since I remember some trickiness about the sign of the proficient and deficient extraction. 

Should be straightforward to extract proficient, deficient, and the whole range, and make sure that proficient + deficient = combined

templates/4b_normalize_image_data.ipynb
```diff
@@ -131,9 +131,18 @@
     "for fov in fovs:\n",
```
ngreenwald, line 131: <p>I think its better to leave this out for now. We shouldn't do any processing of the proficient tifs. They're only there in case something changes in the future. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/342/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

alex-l-kong: <p>We'll also need to remove this call from <code>watcher_callbacks.generate_pulse_heights</code> as well, since it also calls <code>write_mph_per_mass</code> a second time for mass proficient MPH normalization.</p>

src/toffy/watcher_callbacks.py
alex-l-kong: We should add a similar flag to notebook 3b since that also extracts proficient data.

alex-l-kong:
> Looks good. You should also confirm that the offsets are being handled correctly for mass extraction, since I remember some trickiness about the sign of the proficient and deficient extraction.
> 
> Should be straightforward to extract proficient, deficient, and the whole range, and make sure that proficient + deficient = combined

Proficient + deficient unfortunately does not equal combined via the current offset. For most channels, the vast majority of values (about 95-99%) match up, but some channels (ex. Noodle, Fe, Au) have little in common between proficient + deficient and combined. Other channels like H3K27me3 and H3K9ac are also a bit more problematic (70-79%).

In terms of offset, I've been working with:

- Combined from -0.3 to 0.3
- Proficient from 0 to 0.3
- Deficient from -0.3 to 0

ngreenwald:
Hmmm. I can't think of a reason why that would be the case. -0.3 to 0.3 should be the same as the half interval combined. 
Can you look into where the differences are?

ngreenwald:
Mike might have an idea (other than a bug in the code) of how these could be different. Maybe you can ask him at pipeline meeting today?

alex-l-kong:
> Mike might have an idea (other than a bug in the code) of how these could be different. Maybe you can ask him at pipeline meeting today?

Mike thinks the integration ranges are the issue. Both the lower and upper bound on `mibi-bin-tools` are inclusive.

This means that the range goes from `[-0.3, 0]` for deficient, and `[0, 0.3]` for proficient. Mike said this explains why deficient + proficient > total (-0.3, 0.3) and not to worry about it.

ngreenwald:
Okay, will you redo the calculation with -.3 to -.001 and see if that resolves the difference?

alex-l-kong:
@ngreenwald there were no changes to the final results.

I did notice that the Noodle channel had odd panel values specified (116.7, 125.0). In the current mass proficient setup, these do not get corrected since the offset of start is not -0.3. I wonder if at least for Noodle specifically it explains the discrepancy.

ngreenwald:
Okay, then I would say we hold off on merging this until it's clear what the problem is. Is it always the same difference for a given channel in different images, or does it depend? Are all pixels equally affected or only some? How does it differ across different channels?

You can skip extracting the proficient channels for the current mibi runs as well until we figure it out. 

And yes, differences in noodle channel don't matter since it's spanning multiple masses.

alex-l-kong:
Across different channels, the differences are inconsistent for channels between different FOVs, and this varies by channel as well (some channels such as CD14, CD38, PDL1, etc. experience minimal inconsistencies, while others such as Fe, H3K27me3, etc. experience much more significant inconsistencies).

In general, across all images and channels, each region does seem to be equally affected. The average distance between differing pixels between pos + neg images vs full extraction images is around 10 pixels. However, channels such as Fe experience much more frequent clusters of extreme differences between pos + neg vs full extraction images, while others such as CD14 may see frequent clusters of small differences but no clusters of large differences.

This does make sense since Fe is a channel that experienced much more significant inconsistencies in magnitude, whereas CD14 didn't.

alex-l-kong:
> Okay, can you make some summary plots of the differences so we can set a sense of how big a problem this is across different channels?

Yup, will have some ready for tomorrow's standup.

alex-l-kong:
Located one key flaw in the current mass proficient pipeline: the way the `wrapping_functions` work in `mibi_bin_tools.bin_files._set_tof_ranges` which converts mass ranges to time of flight ranges.

By default, the upper_tof_range in this function gets wrapped in an `np.ceil` call, while the lower_tof_range gets wrapped in an `np.floor` call. For deficient extraction, the upper_tof_range is 0 (gets np.ceilinged). However, for proficient extraction, 0 is now the lower_tof_range and thus gets np.floored.

This causes leakage in the time of flight ranges that gets passed to the extraction process. Suppose the full range of integration is [12, 20] for a target. To my knowledge, the deficient range should ideally be [12, 16] for deficient and [16, 20] for proficient. However, what we're seeing is a deficient range of [12, 17] and a proficient range of [16, 20]. This happens because `np.ceil` pushes the cutoff at upper_tof_range=0 up to 17 for deficient, but the cutoff at lower_tof_range=0 down to 16 for proficient. This range leakage also explains why we're seeing negative_ext + positive_ext > full_ext across all channels and FOVs. I confirmed that adjusting the ranges also ensures negative_ext + positive_ext == full_ext.

There are ways to correct for the proficient range window, but I also want to verify whether the deficient time of flight extraction window is correct. @ngreenwald what do you think?

ngreenwald:
Sounds good! Will you put together a brief summary for Friday? Mike will probably be the best to decide how we want to set it up. 

 To confirm, we can now replicate to the full extraction by adding together the partial extraction? Will you check a couple other integration windows as well to see? i.e. -0.3 to -0.1 is the same as -0.3 to -0.2 and -0.2 to -0.1?

alex-l-kong:
@ngreenwald yes, we can now replicate full = sum of neg + pos. I've confirmed this for other ranges as well.

alex-l-kong:
Per discussions with Mike, we'll define proficient and deficient as the following windows:

**Deficient**: (-0.3, -0.06)
**Proficient**: (-0.06, 0.3)

alex-l-kong:
Need to test this PR on the next cohort CIMAC/SPAIN runs before requesting a review.

alex-l-kong:
Per our meeting a few weeks ago, we'll be reverting the definition of deficient and proficient back to:

- Deficient: (-0.3, 0)
- Proficient: (0, 0.3)

pyproject.toml
```diff
@@ -38,7 +38,7 @@ metadata = false
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
 alpineer = ">=0.1.9"
-mibi-bin-tools = "0.2.9"
+mibi-bin-tools = "0.2.10"
```
ngreenwald, line 41: Will this require any changes to the readme, releases, etc?

src/toffy/watcher_callbacks.py
```diff
@@ -159,17 +167,36 @@ def _generate_fov_data(
             replace=replace,
             time_res=time_res,
         )
-
         self.__panel = panel
 
-    def extract_tiffs(self, tiff_out_dir: str, panel: pd.DataFrame, **kwargs):
+        if extract_prof:
+            # modify from (-0.3, 0) to (0, 0.3) for proficient
+            panel_prof = modify_panel_ranges(panel, start_offset=0.3, stop_offset=0.3)
```
ngreenwald, line 174: Why are these the same?

ngreenwald (changes requested):
Looks good, see below

src/toffy/watcher_callbacks.py
alex-l-kong: The offset args define the amount that needs to get added to the Start and Stop columns respectively, not the final offset. So in this case, we need to add 0.3 to the Start column and the Stop column to ensure the range goes from (-0.3, 0) to (0, 0.3).

pyproject.toml
alex-l-kong: We will need to make a new release so that users can have the updated `mibi-bin-tools==0.2.10`.

src/toffy/watcher_callbacks.py
srivarra: I think it would be more intuitive to be able to set the `start_offset = -0.3` and the `stop_offset = 0.0`.

srivarra (changes requested):
Looks good. Just a comment about the setting the ranges.

src/toffy/watcher_callbacks.py
alex-l-kong: I agree it's not the most intuitive, however it will be problematic to use `start_offset=-0.3` and `stop_offset=0.0`, since by default the panel will be specified in the (-0.3, 0) range.

This function is not one that should ever get called by the user. The comment is not that intuitive, so I will adjust it so it's clear that the offsets are added to the `'Start'` and `'Stop'` columns.

pyproject.toml
alex-l-kong: No changes to the readme needed, however.

srivarra (approved):
Wild stuff!

camisowers (approved):

PR #359, opened 2023-04-17
Add functionality to catch file renaming to watcher

### LINKED ISSUE #339: Ensure watcher picks up final `.bin` files and not temp `.bin.{hash}` files
**Describe the bug**

Occasionally, for certain FOVs, two `.bin` files are being written:

- The final `.bin` file
- A temp `.bin` file that gets preceded by a hash (ex. `fov-1-scan-1.bin.aBcDeF`)

The watcher currently doesn't know how to handle this. We've already suppressed the warning that prevents the temp file from throwing warnings, but the watcher doesn't pick up the final `.bin` file, so it hangs without processing it.

The underlying issue is the way the CAC processes these images. Jay from Ionpath says there's no way around it as it's an underlying OS issue.

**Expected behavior**

We should only watch for `.bin` files that are completely suffixed by `.bin` (and not `.bin.{hash}`).

**To Reproduce**

There's seemingly no rhyme or reason as to when this will occur, but it does pop up for some runs.

alex-l-kong:

Closes #339. To address the issue of temp bin files and their later renaming, add watcher functionality that can record these events.

Add an `on_moved` method to `FOV_EventHandler` which makes use of `watchdog.on_moved` events (handled by `FileMovedEvent`). Other than this, the process is exactly the same as `on_created`, so abstract shared functionality to `_run_callbacks`.

`_slow_copy_sample_tissue_data`, the watcher simulator, also includes a process to first generate the temp `.bin` file in `dest`, wait one second, then rename it. This reenacts what actually happens on the CACs.

**Remaining issues**

We will need to test this on an in-process run (and not one that has already been completed).

### DIFF
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -208,7 +208,7 @@ class FOV_EventHandler(FileSystemEventHandler):
             for name in files:
                 self.on_created(FileCreatedEvent(os.path.join(root, name)))
 
-    def _callback_runner(self, event: Union[FileCreatedEvent, FileMovedEvent]):
+    def _run_callbacks(self, event: Union[FileCreatedEvent, FileMovedEvent]):
         # check if what's created is in the run structure
         try:
             fov_ready, point_name = self.run_structure.check_run_condition(event.src_path)
@@ -256,7 +256,7 @@ class FOV_EventHandler(FileSystemEventHandler):
                 file creation event
         """
         super().on_created(event)
-        self._callback_runner(event)
+        self._run_callbacks(event)
 
     def on_moved(self, event: FileMovedEvent):
         """Handles file renaming events
@@ -269,7 +269,7 @@ class FOV_EventHandler(FileSystemEventHandler):
                 file creation event
         """
         super.on_moved(event)
-        self._callback_runner(event)
+        self._run_callbacks(event)
 
     def check_complete(self):
         """Checks run structure fov_progress status
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -256,92 +256,20 @@ class FOV_EventHandler(FileSystemEventHandler):
                 file creation event
         """
         super().on_created(event)
-        # self._callback_runner(event)
+        self._callback_runner(event)
 
-        # check if what's created is in the run structure
-        try:
-            fov_ready, point_name = self.run_structure.check_run_condition(event.src_path)
-        except TimeoutError as timeout_error:
-            print(f"Encountered TimeoutError error: {timeout_error}")
-            logf = open(self.log_path, "a")
-            logf.write(
-                f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- '
-                f"{event.src_path} never reached non-zero file size...\n"
-            )
-            self.check_complete()
-            return
+    def on_moved(self, event: FileMovedEvent):
+        """Handles file renaming events
 
-        if fov_ready:
-            print(f"Discovered {point_name}, beginning per-fov callbacks...")
-            logf = open(self.log_path, "a")
+        If FOV structure is completed, the fov callback, `self.fov_func` will be run over the data.
+        This function is automatically called; users generally shouldn't call this function
 
-            logf.write(
-                f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- Extracting {point_name}\n'
-            )
-
-            # run per_fov callbacks
-            logf.write(
-                f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- '
-                f"Running {self.fov_func.__name__} on {point_name}\n"
-            )
-
-            self.fov_func(self.run_folder, point_name)
-            self.run_structure.processed(point_name)
-
-            if self.inter_func:
-                self.inter_func(self.run_folder)
-
-            logf.close()
-            self.check_complete()
-
-    # def on_moved(self, event: FileMovedEvent):
-    #     """Handles file renaming events
-
-    #     If FOV structure is completed, the fov callback, `self.fov_func` will be run over the data.
-    #     This function is automatically called; users generally shouldn't call this function
-
-    #     Args:
-    #         event (FileCreatedEvent):
-    #             file creation event
-    #     """
-    #     super.on_moved(event)
-    #     self._callback_runner(event)
-
-    #     # # check if what's created is in the run structure
-    #     # try:
-    #     #     fov_ready, point_name = self.run_structure.check_run_condition(event.src_path)
-    #     # except TimeoutError as timeout_error:
-    #     #     print(f"Encountered TimeoutError error: {timeout_error}")
-    #     #     logf = open(self.log_path, "a")
-    #     #     logf.write(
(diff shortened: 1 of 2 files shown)


src/toffy/fov_watcher.py
```diff
@@ -208,21 +213,14 @@ def __init__(
...
-        super().on_created(event)
-
+    def _run_callbacks(self, event: Union[DirCreatedEvent, FileCreatedEvent, FileMovedEvent]):
         # check if what's created is in the run structure
         try:
-            fov_ready, point_name = self.run_structure.check_run_condition(event.src_path)
+            if type(event) in [DirCreatedEvent, FileCreatedEvent]:
+                file_trigger = event.src_path
+            else:
+                file_trigger = event.dest_path
```
srivarra, line 222: So if the original file is called `fov-n-scan-1.bin.asdf`, the destination path would then be `fov-n-scan-1.bin`?

srivarra (approved):
I think it looks good, just wanted some clarification.

tests/fov_watcher_test.py
```diff
@@ -29,24 +29,44 @@
...
+                # copy to a temporary file with hash extension, then move to dest folder
+                new_tissue_path = os.path.join(COMBINED_DATA_PATH, "." + tissue_file + ".aBcDeF")
+                shutil.copy(tissue_path, new_tissue_path)
+                shutil.copy(new_tissue_path, dest)
+                os.remove(new_tissue_path)
+            else:
+                shutil.copy(tissue_path, dest)
+
+            # handle renaming
+            if temp_bin and ".bin" in tissue_file:
```
camisowers, line 66: These are the same conditions as the if statement above right? Could probably move this up instead of re-checking.

src/toffy/fov_watcher.py
alex-l-kong: Yes

tests/fov_watcher_test.py
alex-l-kong: Good idea!

alex-l-kong:
@ngreenwald renaming is handled correctly. One FOV has timed out, but that's was on the CAC backend for being too slow copying one of the JSON files over.

We might need to increase the timeout threshold if this happens more frequently.

ngreenwald (changes requested):
What is the current timeout threshold? The data gets copied over at the top of the hour; if it's in progress at the top of the hour, it needs to wait for the subsequent hour before it will be ready to get processed.

alex-l-kong:
> What is the current timeout threshold? The data gets copied over at the top of the hour; if it's in progress at the top of the hour, it needs to wait for the subsequent hour before it will be ready to get processed.

It's set to just under 62 minutes (1.03 * 60 * 60 seconds). I noticed that for some FOVs it took longer than that for the .bin and .json files to be fully written.

ngreenwald (changes requested):
Okay, lets bump it to 130 minutes. If we still see timeouts, it means something else is going on.

alex-l-kong:
@ngreenwald looking through the code again, another big flaw is `FOV_EventHandler` isn't thread safe. `watchdog` listens for multiple `on_created` and `on_moved` events at the same time, but this will create a race condition if either of these get invoked while another FOV is currently processing. I noticed that FOVs sometimes got processed out of order with the watcher, which seems to confirm this.

In addition to upping the timeout range, I'll also demo a `threading.lock` mechanism on the next run to see if this will help.

alex-l-kong:
@ngreenwald we're no longer timing out nor processing FOVs out of order, but some FOVs have still been skipped.

I monitored the run today and the watcher runs into trouble when a FOVs .bin and .json file get fully written out while another is being processed. The CAC also tends to write .bin files in batches, which only further magnifies the problem. I think the watcher was originally written under the assumption that FOVs would always complete processing before another one gets written. However, this has rarely ended up being the case across an entire run.

There are two solutions we can look into:

1. Maintain a separate class variable to check the last `.bin` file number processed. When an `on_created` or `on_moved` event gets invoked again, modify the watcher to find all the valid `.bin` and `.json` files up to the one just created/moved, then process all of those. Definitely the simplest solution, only other requirement here is we might need to increase the timeout a bit further.
2. Use `asyncio`. This is a cooler solution that could offer a speed up if multiple `.bin` files need processing at once. However, I'd also consider it a slightly riskier solution, especially if multiple threads spawn at once. We could use a semaphore to control the number of concurrent threads to 5, but then we might still have to pick up after `.bin` files that got created while the 5 were processing. Basically, a lot more to consider here, `watchdog` doesn't have great in-built `async` options.

alex-l-kong:
Regardless of the solution we go with, renaming is caught correctly now. Do we want to open a separate PR to address the issue above or continue to work off of this branch?

src/toffy/fov_watcher.py
```diff
@@ -44,18 +52,16 @@ def __init__(self, run_folder: str, timeout: int = 10 * 60):
             if run_order * scan < 0:
                 raise KeyError(f"Could not locate keys in {run_folder}.json")
 
-            fov_names = [f"fov-{run_order}-scan-{s + 1}" for s in range(scan)]
-
-            # identify moly points
+            # scan 2's don't contain significant imaging data per new MIBI specs
+            fov_name = f"fov-{run_order}-scan-1"
```
srivarra, line 56: What information does a scan-2 contain?

src/toffy/fov_watcher.py
```diff
@@ -204,75 +215,234 @@ def __init__(
...
+            return None, None
+
+    def _generate_callback_data(self, point_name: str):
+        print(f"Discovered {point_name}, beginning per-fov callbacks...")
+        logf = open(self.log_path, "a")
+
+        logf.write(f'{datetime.now().strftime("%d/%m/%Y %H:%M:%S")} -- Extracting {point_name}\n')
+
+        # run per_fov callbacks
+        logf.write(
```
srivarra, line 262: Logging sounds great for this.

srivarra (approved):
This looks great!

src/toffy/fov_watcher.py
```diff
@@ -49,18 +52,16 @@ def __init__(self, run_folder: str, timeout: int = 10 * 60):
...
+            self.fov_progress[fov_name] = {"json": False, "bin": False}
 
-            for fov_name in fov_names:
-                self.fov_progress[fov_name] = {
-                    "json": False,
-                    "bin": False,
-                }
+        # compute the highest FOV number, needed for checking if final FOV processed
+        fov_nums = sorted([int(f.split("-")[1]) for f in self.fov_progress.keys()])
+        self.highest_fov = fov_nums[-1]
```
camisowers, line 64: Since we are only adding scan 1's to `fov_progress`, can we simply do `self.highest_fov = len(self.fov_progress.keys())`?

camisowers (changes requested):
Looks good! Just a minor simplicity suggestion.

src/toffy/fov_watcher.py
alex-l-kong: To my knowledge, `scan-2` is generally just Moly points anyways.

camisowers (approved):

ngreenwald (approved):
Great, glad you were able to tease out all these bugs.

PR #361, opened 2023-04-18
Norm workflow

### LINKED ISSUE #325: Default normalization curves
The current pipeline prompts users to generate their own normalization curve. This has caused confusion, and results in an unnecessary additional step. 

Hadeesha is constructing default normalization curves for the instruments. We should add these curves to toffy. There will be 2 curves; one for the default instrument settings, and one for our custom angelolab settings. This should be a flag at the top of the normalization notebook. Based on how this flag is set, that notebook should either use the default, or the angelolab custom curve. The functionality for creating the normalization curve should be removed from notebook 1.

### LINKED ISSUE #344: Update tuning curve process for normalization
**Relevant background**

There's been a bit of confusion over normalization curves since outdated ones have been unknowingly used. This has led to issues such as negative values being returned by the normalization notebook.

To address this, the user should now select from one of two normalization curves, with an option to set which one in the notebook.

**Design overview**

Having dedicated tuning curve files means the process of generating it in the setup notebook should be removed.

In the normalization notebook, the user should be provided a flag to choose which tuning curve they wish to use. This flag will then propagate to the actual normalization process, which then pulls the data from the specified tuning curve file.

The tuning curves themselves should be saved in a dedicated folder for ease of access (and not directly in `src`).

**Code mockup**

_Tuning curve files_

**NOTE: file names STC**

The plan is to have two normalization curves: one for default settings (`default_tuning_curve.json`), and one for custom Angelo Lab settings (`angelolab_tuning_curve.json`).

Notebook 1 originally placed these directly in `src/toffy`. It will be better to put these in a dedicated folder in the root `toffy` directory. This folder should be called `tuning_curves`.

_4b_normalize_image_data.ipynb_

Because there are now two normalization curves to choose, the user will need to specify which one they want to use at the beginning of this notebook.

In the cell where `run_name` and `panel_path` get set, include another variable called `tuning_curve_file` that defaults to `default_tuning_curve.json`. Include in the documentation that the user can change this to `angelolab_tuning_curve.json` if they prefer the custom Angelo Lab tuning curve.

The call to `normalize.normalize_image_data` will need to explicitly specify `norm_func_path=os.path.join(.., tuning_curves, tuning_curve_file)`.

_normalize.normalize_image_data_

The default value for `norm_func_path` should be `os.path.join(.., tuning_curves, 'default_tuning_curve.json')`.

_1_set_up_toffy.ipynb_

The entirety of section 3 (`Generate sensitivity curve`) should be nuked.

_normalize.create_tuning_curve_

This function is solely used by section 3 of notebook 1, so it will also be nuked.

**Required inputs**

The user will need to specify which tuning curve to use in notebook 4b.

**Output files**

Same as before.

**Timeline**
- [ ] A couple days
- [X] A week
- [ ] Multiple weeks. For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

4/13 (dependent on @HPiyadasa in computing the normalization curves, will follow up)

4/17

alex-l-kong:

Closes #325. Closes #344.

See issue template for #344.

**Remaining issues**

Need to document when to use which tuning curve. Still more tuning curves potentially on the way.


srivarra (approved):
LGTM!

src/toffy/normalize.py
```diff
@@ -751,7 +751,7 @@ def normalize_image_data(
     img_sub_folder="",
     mass_obj_func="poly_2",
     extreme_vals=(0.4, 1.1),
-    norm_func_path=os.path.join("..", "toffy", "norm_func.json"),
+    norm_func_path=os.path.join("..", "tuning_curve", "default_tuning_curve.json"),
```
camisowers, line 754: should be "tuning_curves"

camisowers (changes requested):
A few questions about the difference between the curves, but the code looks good! Do we think it's worth adding some of the curve info to [section 4](https://github.com/angelolab/toffy#4-processing-mibi-data) in the readme? Most users probably don't need to know the specifics within the notebook, but it is probably good to have an explanation somewhere with a link to the `generating_tuning_curve.ipynb` for reference.

templates/4b_normalize_image_data.ipynb
```diff
@@ -37,7 +37,8 @@
    "source": [
```
camisowers, line 37: <p>Kind a crowded bullet point. I think we could tweak the explanation a bit.</p><p> </p><p>- <code>tuning_curve_file</code> should point to a tuning curve contained in <code>toffy/tuning_curves</code> (<code>P9_450_norm_func.json</code>, <code>P9_1300_norm_func.json</code>, or <code>P9_2600_norm_func.json</code>)</p><p> </p><p>Include the ionpath default specification in bold underneath.</p><p>**<code>P9_2600_norm_func.json</code> is the curve that should be used when running the instrument with Ionpath's default settings.** <code>P9_450_norm_func.json</code> and <code>P9_1300_norm_func.json</code> are curves with Angelo Lab-specific parameters.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/361/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

camisowers, line 37: <pre class="ql-syntax" spellcheck="false">Line #8.    tuning_curve_file = 'P9_450_norm_func.json'
</pre><p>What's the difference between the two Angelo Lab curves? Why is it defaulted to one over the other? And should we maybe make the default in the notebook be the ionpath P9_2600_norm_func file instead?&nbsp;</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/361/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

alex-l-kong: Definitely agree with making the default 2600.

alex-l-kong: I've added a tidbit to the documentation telling the user to refer to the power supply settings to see which curve is most applicable.

camisowers:
Nice! Looks like there was a weird timeout with the coverage check, but other than that it looks good.

ngreenwald (changes requested):
@HPiyadasa, was P9 the curve that looked weird, or M13? We should include the one that looks the most normal.

alex-l-kong:
@HPiyadasa @ngreenwald I got rid of the explicit documentation for selecting 400, 1300, vs 2600 curves, since these curves are specifically for our lab but other users are likely generating their own curves.

alex-l-kong:
> @HPiyadasa @ngreenwald I got rid of the explicit documentation for selecting 400, 1300, vs 2600 curves, since these curves are specifically for our lab but other users are likely generating their own curves.

Actually I'll add this documentation back in based on a previous discussion. Per today's discussion any of the norm curves at these settings will work (Mike suggests averaging).

ngreenwald:
I agree, let's average them, then name them `avg_norm_func_450`, etc.

templates/generate_tuning_curve.ipynb
```diff
@@ -0,0 +1,136 @@
+{
```
ngreenwald, line 1: <p>Add a preface to this notebook saying that we've generated standard tuning curves which will work across instruments, so unless there is a specific reason the user needs a custom curve, this notebook shouldn't be necessary, use the default ones, etc. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/toffy/pull/361/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

ngreenwald (changes requested):
Looks good, just some more description of the new notebook

ngreenwald (approved):

PR #371, opened 2023-05-09
Next Release - v0.1.3

srivarra:

Updates a few requirements and the `environment.yml` file to support Python 3.9.

Removed the `.coveragerc` file as its contents are already present in the `pyproject.toml` file. Updated the formatting / linting configs for Python 3.9+ support.

**Remaining issues**

N/A.


alex-l-kong (approved):
Nice

environment.yml
```diff
@@ -2,14 +2,15 @@ name: toffy_env
...
   - seaborn
   - scikit-learn
   - jupyterlab
   - jupyter_contrib_nbextensions
   - pip:
-    - toffy
     - natsort
     - watchdog
+    - mibi-bin-tools
+    - toffy==0.1.3
```
camisowers, line 16: We need to pin mibi-bin-tools==0.2.9 and alpineer==0.1.7 right?

camisowers (changes requested):

environment.yml
srivarra: @camisowers The minimum versions are set in the `pyproject.toml` file. It isn't necessary to pin them in the environment.yml. Would it be better to pin them in the `environment.yml`

camisowers: Isn't it going to keep doing the thing where it pip installs the latest version for the conda environment? 

I think for people using toffy_env it needs to be explicitly pinned. Really I'm inclined to say they should be explicitly pinned in `pyproject.toml` as well, since the multiple tiled stitching was able to pass testing when it would actually error out for toffy_env users.

srivarra: Sure, sounds good. I'll pin it in both.

camisowers (approved):

PR #374, opened 2023-05-17
Add a scale factor for `image_stitching.stitch_images` for Photoshop compatibility

### LINKED ISSUE #355: create_stitched_images callback produces tiffs that are incompatible with photoshop

**Describe the bug**
The function to generate stitched images makes TIFs that can be opened in imageJ, but show up as blank in photoshop. This is usually because of an issue with the dtype of the tif being saved. 

**Expected behavior**
TIFs that can be opened in photoshop.

**To Reproduce**
Run the stitched images notebook.

alex-l-kong:

Closes #355. The current stitched images are not viewable in Photoshop due to scaling issues, so we need to rescale prior to saving.

In Rosetta, we scale each image down by 200 to fix this. This should also be added to the `stitch_images` function:

```
current_img = stitched.loc["stitched_image", :, :, chan].values / scale
```

`scale` is a param that gets passed to `stitch_images`. As with Rosetta, it defaults to 200.

**Remaining issues**

The extracted images themselves have the same scaling issue, but I think it's best to leave these because they get scaled later on in Rosetta.

### DIFF
--- a/src/toffy/image_stitching.py
+++ b/src/toffy/image_stitching.py
@@ -193,7 +193,7 @@ def stitch_images(
                 )
                 fname = os.path.join(tile_stitched_dir, chan + "_stitched.tiff")
                 stitched = data_utils.stitch_images(image_data, num_cols)
-                current_img = stitched.loc["stitched_image", :, :, chan].values
+                current_img = stitched.loc["stitched_image", :, :, chan].values / scale
                 image_utils.save_image(fname, current_img)
 
         else:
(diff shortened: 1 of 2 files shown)


camisowers:
> Is there a way to test for this?

The stitched data is already loaded in to test the shape is correct, can just check that the values in the array are scaled down
(see [here](https://github.com/angelolab/toffy/blob/f0de684435ba8c6eb324b228daa8d65fd045101c/tests/image_stitching_test.py#L155)).

srivarra (approved):
Looks good!

ngreenwald (approved):
Looks good

PR #383, opened 2023-06-27
Allow for image specified for flatfield correction to be empty

### LINKED ISSUE #299: Issue with ffc_masses argument in rosetta.compensate_image_data
In compensate_image_data, the default input for ffc_masses is ffc_masses=[39]. If you do not want to apply flat field correction you cannot provide ffc_masses=None because in line 223 the function expects to iterate over ffc_masses to produce ffc_channels. Since NoneType is not iterable, this throws an error. If you instead provide an empty list (ie. ffc_masses = []) then you make it past line 223, but when you get to line 267 the function states: 

`        # apply flat field correction if specified
        if ffc_channels is not None:
            misc_utils.verify_in_list(flat_field_correction_masses=ffc_channels,
                                      all_masses=acquired_targets)`

There are two issues here: 

1.  There is no boolean argument to specify ffc like there is for streak correct (correct_streaks). So there is no place the user 'specifies' this processing step
2. If you get to this stage by feeding an empty list to ffc_masses, then ffc_channels an error occurs when the verify_in_list function attempts to index on the empty list. 

My suggestion would be to add a boolean argument to specify in ffc should be done. If this is false, set ffc_channels = None, otherwise set ffc_channels equal to the existing argument in line 223. This way the script runs without issue. This is how I worked around the problem.

cliu72:

Handle the case where the image for flatfield correction is empty, and allow for the user to specify no images for flatfield correction. And added a missing space in the jupyter notebook.

1. In the flatfield correction function, return an empty array if the input array is all zeros.
2. Allow the user to specify "None" for flatfield correction masses.

Closes #299 
**Remaining issues**

None

### DIFF
--- a/tests/rosetta_test.py
+++ b/tests/rosetta_test.py
@@ -655,7 +655,7 @@ def create_rosetta_comp_structure(
     batch_size=1,
     gaus_rad=1,
     norm_const=200,
-    ffc_channels=["chan_39"],
+    ffc_masses=[39],
     correct_streaks=False,
     streak_chan="Noodle",
 ):
--- a/tests/rosetta_test.py
+++ b/tests/rosetta_test.py
@@ -227,7 +227,7 @@ def test_flat_field_correction():
     assert not np.array_equal(corrected_img, input_img)
 
     # test empty image
-    with pytest.warns(UserWarning, match=''):
+    with pytest.warns(UserWarning, match='Image for flatfield correction is empty'):
         input_img = np.zeros((10, 10))
         corrected_img = rosetta.flat_field_correction(img=input_img)
 
--- a/tests/rosetta_test.py
+++ b/tests/rosetta_test.py
@@ -226,6 +226,13 @@ def test_flat_field_correction():
     assert corrected_img.shape == input_img.shape
     assert not np.array_equal(corrected_img, input_img)
 
+    # test empty image
+    with pytest.warns(UserWarning, match=''):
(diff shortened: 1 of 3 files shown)


tests/rosetta_test.py
```diff
@@ -226,6 +226,13 @@ def test_flat_field_correction():
     assert corrected_img.shape == input_img.shape
     assert not np.array_equal(corrected_img, input_img)
 
+    # test empty image
+    with pytest.warns(UserWarning, match="Image for flatfield correction is empty"):
+        input_img = np.zeros((10, 10))
+        corrected_img = rosetta.flat_field_correction(img=input_img)
+
+        assert not np.any(input_img)
```
ngreenwald, line 234: You probably meant to check the `corrected_img`, not the `input_img`. Although I suppose they're both equal to zero.

tests/rosetta_test.py
```diff
@@ -648,7 +655,7 @@ def create_rosetta_comp_structure(
     batch_size=1,
     gaus_rad=1,
     norm_const=200,
-    ffc_channels=["chan_39"],
+    ffc_masses=[39],
```
ngreenwald, line 658: I would add a test that runs the top-level function with `ffc_masses=None` to ensure that part works as well.

tests/rosetta_test.py
cliu72: Whoops my b, fixed

ngreenwald (approved):

PR #386, opened 2023-06-30
Turn on y ticks for QC plots

### LINKED ISSUE #378: qc_metrics plots don't have axis labels

**Describe the bug**
The current version of the watcher on Drago produces QC plots as expected, but there are no axis labels anywhere on the plot. I think it's okay to omit x axis labels if that makes things look better, but we need to keep the y axis labels, otherwise the plot is hard to interpret

image

alex-l-kong:

Closes #378. QC plots need to have corresponding y-ticks because the scaling for each channel is different.

Remove any setting in `visualize_qc_metrics` where `yticks` or `yticklabels` are set to `[]`.


srivarra (approved):

camisowers (approved):

ngreenwald (changes requested):
What does the output look like now?

alex-l-kong:
@ngreenwald here's an example plot:

Non-zero mean intensity_barplot_stats

PR #390, opened 2023-07-12
Make sure watcher logging actually writes to file on CACs

alex-l-kong:

It appears that on the CACs, the logging module isn't consistently writing to files. This will make it difficult to debug.

Explicitly specify the level of the `logging` module to `logging.INFO` so it can capture both `.info` and `.warning`.


srivarra (approved):
Looks good, what do you know what logging package watchdog uses?

alex-l-kong:
> Looks good, what do you know what logging package watchdog uses?

Don't think `watchdog` has logging built in to it, I've just been using the `logging` library.

camisowers (approved):
Sounds good to me.

PR #391, opened 2023-07-17
FOV resolution check 

### LINKED ISSUE #388: Identify resolution per image
Now that we're going to be imaging with different combinations of pixel resolution and image size, the change of an error when setting this up increases. 

A QC check that will look at the number of pixels and FOV size for each FOV, compute the resolution for each image, then print this out for all the FOVs in a run.

camisowers:

Closes #388. Calculates the image resolution for each FOV using `frameSizePixels[width]`/`fovSizeMicrons` and outputs them for the user to check.

Add a function `check_fov_resoltutions()` which retrieves the information from the run file and calculated the resolution for each FOV. 

Print resolutions in the notebook and have a save_path arg if the user would like to specify a file to save to.
Example notebook output:

Screenshot 2023-07-19 at 9 52 37 AM

**Remaining issues**

- [x] Should we leave this as its own notebook or throw the check in somewhere in the pipeline?


src/toffy/json_utils.py
srivarra, line 1: Could this work with notebook 5: rename and reorganize? Or is the csv created not intended to be used in downstream analysis/

src/toffy/json_utils.py
```diff
@@ -194,3 +195,40 @@ def check_for_empty_files(bin_file_dir):
...
+
+    fov_names, resolutions = [], []
+    for fov in run_metadata.get("fovs", ()):
+        # get fov names
+        fov_number = fov.get("runOrder")
+        name = f"fov-{fov_number}-scan-1"
+        fov_names.append(name)
+
+        # retrieve pixel and micron specs
+        fov_pixel_length = fov.get("frameSizePixels")["width"]
```
srivarra, line 222: Is there any instance where the wouldn't have this property? If so we could add a default value as a second parameter to `.get()`.

srivarra (changes requested):
Looks good, just a couple usability questions!

src/toffy/json_utils.py
camisowers: I *think* this is supposed to be run early (pre-rosetta). Having FOVs with different resolutions is not supposed to happen, but with Albert's new tool it could come up via accidental user error.

src/toffy/json_utils.py
```diff
@@ -194,3 +195,40 @@ def check_for_empty_files(bin_file_dir):
...
+    # read in run metadata
+    run_file_path = os.path.join(bin_file_dir, run_name + ".json")
+    io_utils.validate_paths([run_file_path])
+    run_metadata = read_json_file(run_file_path, encoding="utf-8")
+
+    fov_names, resolutions = [], []
+    for fov in run_metadata.get("fovs", ()):
+        # get fov names
+        fov_number = fov.get("runOrder")
+        name = f"fov-{fov_number}-scan-1"
```
alex-l-kong, line 218: This would cause an issue if `run_metadata` contains `scan-2` or `scan-3`. In that case is you'd gets something like this:

```
...
fov-n-scan-1: xxx
fov-n-scan-1: xxx # actually fov-n-scan-2's resolution
fov-n-scan-3: xxx # actually fov-n-scan-3's resolution
...
```

The watcher script right now has a similar issue with `runCount` during initialization, however all it does is overwrite static dict keys during initialization, so I left it be. In this case, however, I'd explicitly skip anything that has `scanCount` (or whatever the key is) greater than 1 (or warn the user that anything with scans > 1 will be ignored during extraction).

alex-l-kong (changes requested):
I'd say it's better to put this FOV resolution check directly after the `3a` and `3b` notebooks, since we'd want to verify that the resolution is correct right after image extraction.

If you're really feeling it, you could potentially even add it to the `fov_watcher` (full pipeline) and `bin_extraction` (just extraction) as well.

src/toffy/json_utils.py
camisowers: I had thought that multiple scan info is still stored under a single FOV key in the run metadata? If so, that shouldn't be a problem here, but if the run file structure has changed then yeah this could be a problem.

camisowers: Yup just verified with a run file from 7/05 that there is only a single `runOrder`, `frameSizePixels`, and `fovSizeMicrons` key even when there are multiple scans.

camisowers:
> I'd say it's better to put this FOV resolution check directly after the `3a` and `3b` notebooks, since we'd want to verify that the resolution is correct right after image extraction.
> 
> If you're really feeling it, you could potentially even add it to the `fov_watcher` (full pipeline) and `bin_extraction` (just extraction) as well.

Agreed, will discuss at pipeline today.

src/toffy/json_utils.py
```diff
@@ -194,3 +195,44 @@ def check_for_empty_files(bin_file_dir):
...
```
ngreenwald, line 219: Will you add the user-defined name here as well?

ngreenwald (changes requested):
Looks good, one addition

ngreenwald (approved):

PR #392, opened 2023-07-17
Ensure incompletely extracted bin files get re-extracted

### LINKED ISSUE #389: Extracted images have weird artifact

**Describe the bug**
A subset of the extracted images have a weird image artifact, which to me looks like the pixels are being offset from their true location. This is not present in the images on the ionpath tracker, so it's something about our extraction/saving that isn't working. 

image

**Expected behavior**
Successful extraction, or an error message

**To Reproduce**
Please either copy/paste or screenshot all of the code you ran which produced the error. In addition, copy the full error message.

alex-l-kong:

Closes #389. In some cases, even if the `.json` file is written for a FOV during a MIBI run, the `.bin` file will not have been fully copied over. The watcher needs to account for these cases because otherwise, the data extracted from the `.bin` file will be incomplete, leading to the artifacts highlighted in the issue.

When the control statement in `check_complete` runs, one final pass of all the FOVs needs to be conducted. Offending `.bin` files will have their created timestamp be later than their modified timestamp, in which case, the `.bin` file is once again thrown into the extraction pipeline.

**Remaining issues**

Full testing needs to be conducted on @ngreenwald's cohort.


alex-l-kong:
@ngreenwald with this PR, your STANFORD cohort stitched images don't show the corrupted artifacts on the bottom of images anymore.

tests/fov_watcher_test.py
```diff
@@ -64,9 +72,22 @@ def _slow_copy_sample_tissue_data(
...
+    bin_files = [bfile for bfile in sorted(os.listdir(COMBINED_DATA_PATH)) if ".bin" in bfile]
+
+    # simulate updating the creation time for some .bin files, this tests _check_bin_updates
+    for i, bfile in enumerate(bin_files):
+        if i % 2 == 0:
+            shutil.copy(
+                os.path.join(COMBINED_DATA_PATH, bfile), os.path.join(dest, bfile + ".temp")
+            )
+            os.remove(os.path.join(dest, bfile))
+            os.rename(os.path.join(dest, bfile + ".temp"), os.path.join(dest, bfile))
```
camisowers, line 89: This should trigger `_check_bin_updates()`, but do you check anywhere that this is actually executed? Since the first extraction already generates an output tiff (albeit an incorrect one), I think you need an additional check for the new function beyond the existence of the output images.

For example you could test that the console does in fact print `f"Re-extracting incompletely extracted FOV {fov}"`.

camisowers (changes requested):
Just an additional test suggestion.

tests/fov_watcher_test.py
alex-l-kong: Good catch.

srivarra (changes requested):
Just some docs. Looks good otherwise.

camisowers (approved):

ngreenwald (approved):
Nice stuff

PR #393, opened 2023-07-19
Turn off `combined.csv` file warnings for intermediate callbacks in watcher

### LINKED ISSUE #387: `combine_run_metrics` throws removed combined file warning on each FOV run through the watcher

Each time a new FOV is detected by the watcher and the MPH pipeline runs, the _combined.csv file gets removed and regenerated, but this also comes with a warning that indicates this happens. If the run contains a substantial amount of FOVs, this warning will pop up for each one except the first (with 50, this can get very wordy).

There should at the very least be a flag that turns this off for the watcher. We could consider using `verbose` or something else.

Honestly, I wouldn't be against just removing the warning entirely either.

alex-l-kong:

Closes #387. The new watcher setup means we'll be defaulting to using intermediate callbacks for plotting QC and MPH metrics. This means `combined.csv` files, normally removed at the end of the watcher, now get removed after every FOV. The warning that gets thrown as a result can appear several times on cohorts with many FOVs, which we want to turn off in this situation.

Pass a `warn_overwrite` flag to the relevant functions. Also include this check for QC combined file generation and overwriting.

**Remaining issues**

Need testing on a live run, per the usual.


src/toffy/watcher_callbacks.py
```diff
@@ -358,16 +363,6 @@ def build_callbacks(
...
-    for run_cb in callbacks_with_prereq:
-        argnames = inspect.getfullargspec(getattr(RunCallbacks, run_cb))[0]
-        argnames = [argname for argname in argnames if argname != "self"]
-
-        misc_utils.verify_in_list(required_arguments=argnames, passed_arguments=list(kwargs.keys()))
-
-    callbacks_with_prereq = (
-        run_callbacks + intermediate_callbacks if intermediate_callbacks else run_callbacks[:]
-    )
-
```
camisowers, line 370: Can you explain why this is this getting removed?

camisowers (changes requested):
Looks good, just a question about the removed `build_callbacks` code.

src/toffy/watcher_callbacks.py
alex-l-kong: This particular section was duplicated, so it should be removed.

srivarra (approved):
Looks good!

camisowers (approved):

ngreenwald (approved):

PR #394, opened 2023-07-19
Skip watcher callbacks for FOVs that already have data generated

### LINKED ISSUE #346: Allow watcher to skip previously processed files
If the watcher is restarted on a given run directory, it will reprocess all of the data. 

Like the 3_letter notebooks, it would be convenient if it skipped the previously processed FOVs

**Describe alternatives you've considered**
If the watcher becomes sufficiently stable that it never needs to be restarted, this will be less of an issue. Right now, the intermittent freezing due to fov-x-scan-1-bin.a35dj files means the only way to get out of the stuck loop is to kill the process. However, re-running it isn't feasible since it will start from the beginning.

alex-l-kong:

Closes #346. If the watcher has to be stopped for some reason, currently all the FOVs need to be re-run regardless of whether they were already processed. Fixes this so the watcher skips these FOVs instead.

Modify `watcher_callbacks.py` so that each callback first checks if the relevant files have been written. If so, simply return out. All `fov_watcher.py` will see is that these individual callbacks finished, thus allowing it to update properly.

**Remaining issues**

Testing on a live cohort.


src/toffy/watcher_callbacks.py
```diff
@@ -179,6 +180,11 @@ def extract_tiffs(self, tiff_out_dir: str, panel: pd.DataFrame, **kwargs):
         if not os.path.exists(tiff_out_dir):
             os.makedirs(tiff_out_dir)
 
+        extracted_img_dir = os.path.join(tiff_out_dir, self.point_name)
+        if os.path.exists(extracted_img_dir):
+            warnings.warn(f"Images already extracted for FOV {self.point_name}")
+            return
+
```
camisowers, line 187: Is there any reason to believe that a FOV could be interrupted during the extraction step resulting in missing channel tiffs? If not, this is fine.

alex-l-kong: It's hard to predict where in the extraction process for a FOV the run could be interrupted. I think it's safest to include a test for all cases just in case we don't leave behind any missing pieces.

src/toffy/watcher_callbacks.py
```diff
@@ -242,6 +257,11 @@ def generate_mph(self, mph_out_dir, **kwargs):
         if not os.path.exists(mph_out_dir):
             os.makedirs(mph_out_dir)
 
+        mph_pulse_file = os.path.join(mph_out_dir, f"{self.point_name}" + "-mph_pulse.csv")
```
srivarra, line 260: We can remove the `+` and add the `"-mph_pulse.csv"` to the f-string itself.

src/toffy/watcher_callbacks.py
```diff
@@ -267,6 +287,11 @@ def generate_pulse_heights(self, pulse_out_dir: str, panel: pd.DataFrame = None,
         if not os.path.exists(pulse_out_dir):
             os.makedirs(pulse_out_dir)
 
+        pulse_height_file = os.path.join(pulse_out_dir, f"{self.point_name}" + "-pulse_heights.csv")
```
srivarra, line 290: Same here

srivarra (changes requested):
Looks good, just a small suggestions with f-strings.

srivarra (approved):
Looks good!

src/toffy/watcher_callbacks.py
ngreenwald: So then this needs to be updated then, right? This will skip extraction for all channels even if only one tiff was generated

ngreenwald (changes requested):

ngreenwald (approved):

PR #398, opened 2023-07-25
Fix rescale_raw_images error

### LINKED ISSUE #310: When possible make functions check if exist = skip, if not = run, rather than producing error
Many functions in toffy and ark repo already check for existing folders before running, however there are still functions that error out rather than skipping sections that have already been run previously.

 rescale_raw_imgs function errors out if rescaled folder already exists

Screenshot 2023-01-27 at 11 29 55 AM

camisowers:

Closes #310. If the cell in the rosetta notebook containing `rosetta.rescale_raw_images()` is run more than once, it will give an error messages that is confusing for the user. We instead want to simply skip any images that have already been rescaled and saved the the subfolder.

1. Only create the rescaled subdir if it doesn't exist. 
2. Check for existing channel images in the directory before repeating the process of loading in, scaling, and saving all the images.
3. Update the tests to check the files are not being overwritten when the function is called again.

**Remaining issues**

N/A

### DIFF
--- a/src/toffy/rosetta.py
+++ b/src/toffy/rosetta.py
@@ -677,14 +677,16 @@ def rescale_raw_imgs(img_out_dir, scale=200):
         fov_dir = os.path.join(img_out_dir, fov)
         # create subdirectory for the new images
         sub_dir = os.path.join(fov_dir, "rescaled")
-        os.makedirs(sub_dir)
+        if not os.path.exists(sub_dir):
+            os.makedirs(sub_dir)
         chans = io_utils.list_files(fov_dir)
         # rescale each channel image
         for chan in chans:
-            img = io.imread(os.path.join(fov_dir, chan))
-            img = (img / scale).astype("float32")
             fname = os.path.join(sub_dir, chan)
-            image_utils.save_image(fname, img)
+            if not os.path.exists(fname):
+                img = io.imread(os.path.join(fov_dir, chan))
+                img = (img / scale).astype("float32")
+                image_utils.save_image(fname, img)
 
 
 def generate_rosetta_test_imgs(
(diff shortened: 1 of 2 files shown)


srivarra (approved):
Looks good!

src/toffy/rosetta.py
```diff
@@ -677,14 +677,16 @@ def rescale_raw_imgs(img_out_dir, scale=200):
...
+        if not os.path.exists(sub_dir):
+            os.makedirs(sub_dir)
         chans = io_utils.list_files(fov_dir)
         # rescale each channel image
         for chan in chans:
-            img = io.imread(os.path.join(fov_dir, chan))
-            img = (img / scale).astype("float32")
             fname = os.path.join(sub_dir, chan)
-            image_utils.save_image(fname, img)
+            if not os.path.exists(fname):
```
alex-l-kong, line 686: Per #310 looks like the solution @HPiyadasa wanted writes the `.tiff.` regardless of if a channel file already exists. This makes sense because if the user specifies an existing `sub_dir`, most likely they want to overwrite the existing data there.

alex-l-kong (changes requested):
One clarification related to @HPiyadasa's method in #310

src/toffy/rosetta.py
camisowers: If you look at his code, the solution only writes anything if the "rescaled" subdir doesn't exist in the FOV folder. All reading and writing is done under the statement (`if not os.path.exists(subdir)`), so it's not possible to overwrite.

It really depends on the functionality that makes the most sense for rosetta, but I can't think of what could change about the extracted images that would cause a user to need to re-scale.

camisowers: I basically just added a sanity check to that verifies all the channels are actually in the subdir, rather than skipping that FOV entirely.

alex-l-kong (approved):

ngreenwald (approved):

PR #401, opened 2023-07-31
Ensure watcher can wait on run folder to appear

### LINKED ISSUE #377: Ensure watcher can run even if run folder not written to Data

Currently, the watcher can only start if the folder for a particular run has been written to the MIBIData folder, which only happens after the first FOV is generated. The watcher should run even if this folder has not been created.

Provide a separate listener to watch for the creation of the run folder, then spin up the watcher instance once this gets created.

alex-l-kong:

Closes #377. The watcher is not currently designed to run until the run folder appears. This is problematic because this doesn't happen until about 15-30 minutes into the start of a run. The user needs to be able to spin up the watcher immediately after the run starts, otherwise it causes a significant inconvenience.

`fov_watcher.start_watcher` will now take a new argument, `run_folder_timeout`, that controls how long the watcher waits for the run folder to appear. This allows implementation of timing logic similar to the timeout logic for waiting on `.bin` and `.json` files. The watcher will wait until either the run folder appears (in which case it listens for files inside run folder) or `run_folder_timeout` is reached (in which case it errors out with an informative message to the user).

Occasionally, what happens is the CACs will change the casing and/or add extraneous characters (notably `"T"` as a timestamp marker) to run names. If the run folder specified by the user hasn't already written to the CAC, warn the user to double check the run folder doesn't already exist in the `D:\\Data` folder under a slightly different name.

**Remaining issues**

Should we instead include the warning in a cell prior?

### DIFF
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -560,7 +560,7 @@ def start_watcher(
             "before trying again."
         )
 
-    # allow the watcher to
+    # allow the watcher to poll the run folder until it appears or times out
     run_folder_wait_time = 0
     while not os.path.exists(run_folder) and run_folder_wait_time < run_folder_timeout:
         time.sleep(run_folder_timeout / 10)
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -554,7 +554,7 @@ def start_watcher(
 
     if run_folder_wait == run_folder_timeout:
         raise FileNotFoundError(
-            f"run_folder {run_folder} does not exist. First double check that the data doesn't "
+            f"Timed out waiting for {run_folder}. First double check that the data doesn't "
             "already exist in D:\\\\Data (sometimes, the CACs will change capitalization or add "
             "extra characters to the run name). If that isn't the case, then wait 15-30 minutes "
             "before trying again."
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -546,6 +546,12 @@ def start_watcher(
         zero_size_timeout (int):
             number of seconds to wait for non-zero file size
     """
+    confirm = input(
+        f"Move forward with specified run folder {run_folder}? Double check spelling and "
+        "capitalization, sometimes the CACs will add additional characters or add "
+        "weird casing changes"
+    )
+
     # allow the watcher to
     run_folder_wait_time = 0
     while not os.path.exists(run_folder) and run_folder_wait_time < run_folder_timeout:
--- a/src/toffy/fov_watcher.py
+++ b/src/toffy/fov_watcher.py
@@ -547,12 +547,12 @@ def start_watcher(
             number of seconds to wait for non-zero file size
     """
     # allow the watcher to
-    run_folder_wait = 0
-    while not os.path.exists(run_folder) and run_folder_wait < run_folder_timeout:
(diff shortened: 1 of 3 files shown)


src/toffy/fov_watcher.py
```diff
@@ -540,12 +541,38 @@ def start_watcher(
...
+    if not os.path.exists(run_folder):
+        warnings.warn(
+            f"Waiting for {run_folder}. Please first double check that your run data "
+            "doesn't already exist under a slightly different name in D:\\Data. "
+            "Sometimes, the CACs change capitalization or add extra characters to the run folder. "
+            "If this happens, stop the watcher and update the run_name variable in the notebook "
+            "before trying again."
+        )
+
+    # allow the watcher to
```
srivarra, line 563: The suspense! what is it allowed to do? 🤔

tests/fov_watcher_test.py
```diff
@@ -127,6 +127,103 @@ def test_run_structure(run_json, expected_files, recwarn):
...
+        # ensure warn_overwrite set to False if intermediate callbacks set, otherwise True
+        kwargs["warn_overwrite"] = True if int_cbs else False
+
+        run_folder = os.path.join(tmpdir, "test_run")
+        log_out = os.path.join(tmpdir, "log_output")
+        fov_callback, run_callback, intermediate_callback = build_callbacks(
+            run_cbs, int_cbs, fov_cbs, **kwargs
+        )
+
+        with Pool(processes=4) as pool:
```
srivarra, line 189: Any particular reason to use 4 processes?

srivarra (changes requested):
Looks good, have you tried this on a run yet?

src/toffy/fov_watcher.py
alex-l-kong: I'm giving the developer an opportunity to use their imagination 🤣

tests/fov_watcher_test.py
alex-l-kong: Consistency with the main watcher test below. We need multiple to ensure these functions can run in parallel.

alex-l-kong:
> Looks good, have you tried this on a run yet?

Yup, I tried it on Mako's

tests/fov_watcher_test.py
```diff
@@ -127,6 +127,103 @@ def test_run_structure(run_json, expected_files, recwarn):
...
+                            intermediate_callback,
+                            10,
+                            1,
+                            SLOW_COPY_INTERVAL_S,
+                        ),
+                    )
+
+                    res_scan.get()
+            else:
+                res_scan = pool.apply_async(
```
camisowers, line 210: Just for test timing, I'm wondering if we need to see this watcher test to completion? Or is it sufficient to just exit out of start_watcher once the run folder has been created successfully and the script has successfully began to create outputs from the first few callbacks? 
If the watcher tests don't take that long then this is probably fine as is.

camisowers (commented):
Looks good, just a small test efficiency thought. Also the builds/tests are lagging below.

tests/fov_watcher_test.py
srivarra: Looks like the workflow isn't starting the test process, but I am able to manually run it with.

srivarra (approved):
Looks Good!

ngreenwald (approved):

PR #406, opened 2023-08-02
Ensure runs below fovs_per_num threshold get skipped during Rosetta t…

### LINKED ISSUE #309: 4a_compensate notebook - copy_image_files function - skip over runs that dont have minimum # of FOVs
My current cohort has >30 runs and >1500 FOVs. Therefore I used list_folders function to generate a list of run names to use as I wanted to randomly select 2 FOVs from each run to build a test set of 60 FOVs to make it representative. However, there are runs that failed after the first point was acquired so these runs only have 1 point in them. So I get the following error when trying to create my test set.  

"The run folders ['2022-12-12_Stanford_TA550_Immune', '2022-12-12_Stanford_TA550_Immune_run2', '2022-12-15_Stanford_TA550_Immune'] do not contain the minimum amount of FOVs (2) defined by the fovs_per_run given."

Skip over runs that does not meet min # of FOV requirement but still print those for users to be aware of which runs were omitted when creating the test set. 

**Describe alternatives you've considered**
Generate interactive list to select and deselect runs included for generating the test set

### LINKED ISSUE #340: Print out FOV name in compensation notebook
Currently in [4a_compensate_image_data.ipynb](https://github.com/angelolab/toffy/blob/main/templates/4a_compensate_image_data.ipynb), the cell running the compensation prints out each time an image is processed, so the user can see it is running smoothly. 

Screenshot 2023-04-06 at 3 43 20 PM

However, the images do not process in an order the user can follow easily and so if something goes wrong, it's difficult to pinpoint which set of images potentially caused the issue. 

In [rosetta.compensate_image_data](https://github.com/angelolab/toffy/blob/a7e67d7ee7975a219a0f16f800b7a34c6ae3b72e/src/toffy/rosetta.py#L211), change the following part of the code.
```
for i in range(0, len(fovs), batch_size):
       # OLD
        print("Processing image {}".format(i + 1))

       # NEW
        print("Processing {}".format(fov))

        # load batch of fovs
        batch_fovs = fovs[i : i + batch_size]
        batch_data = load_utils.load_imgs_from_tree(
            data_dir=raw_data_dir,
            fovs=batch_fovs,
            channels=acquired_targets,
            img_sub_folder=raw_data_sub_folder,
        )
```

**Describe alternatives you've considered**
Let the suffering continue. 

**Additional context**
Can probably just be added to #319.

alex-l-kong:

Closes #309. Closes #340. During Rosetta test set generation, sometimes the user will specify a `fovs_per_num` threshold that a few runs don't meet. This will cause the whole process to error out when in reality its fine that those runs don't make it in. If there are a lot of runs (sometimes over 100), it is cumbersome for the user to pick out which ones need to be removed.

To make things simple, just ensure that any runs that don't meet the `fovs_per_num` threshold in `copy_image_files`. Only if none of them do should we error out.

Modify `copy_image_files` so that only a `warnings.warn` is thrown if some FOVs get skipped over. However, do keep the `ValueError` in case no runs get copied over.

**Remaining issues**

This doesn't exactly solve the issue of what happens if the user only ends up with a few valid runs out of several hundred due to a `fovs_per_run` that was unusually high. We might want to specify a percentage threshold for erroring out to account for this (ex. if fewer than 50% of `run_names` meet the threshold, error out and let the user know that the Rosetta results will not be indicative of the desired test set).

### DIFF
--- a/src/toffy/rosetta.py
+++ b/src/toffy/rosetta.py
@@ -634,18 +634,30 @@ def copy_image_files(
         # check number of fovs in each run
         if len(fovs_in_run) < fovs_per_run:
             small_runs.append(run)
+
+    run_names_process = run_names[:]
     if len(small_runs) > 0:
-        raise ValueError(
-            f"The run folders {small_runs} do not contain the minimum amount of FOVs "
-            f"({fovs_per_run}) defined by the fovs_per_run given."
+        # edge case if none of the runs in run_names have length equal to fovs_per_run
+        if len(small_runs) == len(run_names_process):
+            raise ValueError(
+                f"None of the runs specified contain the minimum amount of FOVs ({fovs_per_run}) "
+                "defined by the fovs_per_run given"
+            )
+
+        # warn user that runs below fovs_per_run threshold will be skipped
+        warnings.warn(
+            "The following runs will be skipped because they do not contain the minimum amount "
+            f"of FOVs ({fovs_per_run}) defined by the fovs_per_run given: {small_runs}."
         )
+
+        run_names_process = list(set(run_names_process).difference(small_runs))
 
     # make rosetta testing dir and extracted images subdir
     cohort_rosetta_dir = os.path.join(rosetta_testing_dir, cohort_name)
     os.makedirs(os.path.join(cohort_rosetta_dir, "extracted_images"))
 
     # randomly choose fovs from a run and copy them to the img subdir in rosetta testing dir
-    for i, run in enumerate(ns.natsorted(run_names)):
+    for i, run in enumerate(ns.natsorted(run_names_process)):
         run_path = os.path.join(extracted_imgs_dir, run)
 
         fovs_in_run = io_utils.list_folders(run_path, substrs="fov")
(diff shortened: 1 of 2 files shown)


camisowers (changes requested):
Looks good. Do you want to close #340 with this PR as well? I think I suggested it be handled by the rosetta v2 PR but that one seems to be held up.

alex-l-kong:
> Looks good. Do you want to close #340 with this PR as well? I think I suggested it to be handled by the rosetta v2 PR but that one seems to be held up.

Yeah I can address that one too.

src/toffy/rosetta.py
```diff
@@ -301,7 +301,7 @@ def compensate_image_data(
 
     # loop over each set of FOVs in the batch
     for i in range(0, len(fovs), batch_size):
-        print("Processing image {}".format(i + 1))
+        print(f"Processing FOV {fovs[i]}")
```
camisowers, line 304: Can probably just be "Processing fov-1-scan-1" instead of "Processing FOV fov-1-scan-1".

camisowers (approved):

ngreenwald (approved):
Looks fine, I don't think we need to keep track of how many runs are skipped since it's printed out for the user.

PR #410, opened 2023-08-07
Check for missing fovs after run is complete

### LINKED ISSUE #399: Check for missing FOVs
Sometimes FOVs will fail to be generated due to instrument issues. This is often caught at the time, but sometimes goes unnoticed. 

We should check for list of expected FOVs from the run JSON file, and make sure each of those files exists. This helper function can get called by the watcher as a run callback, and also added to notebook 3f.

camisowers:

Closes #399. Checks that bin/json files are generated for all FOVs outlined in the run file.

Also addresses the watcher edge case where a run crashes or does not output the bin/json files of one of the fovs in the run file. Currently the watcher notebook will hang forever and ever, since our current FOV timeout system is based on how long it takes for the full bin file to be completely generated (only triggered once it is written with size zero).

Create a helper function to add to notebook 3f and also implement a run callback in watcher. 

Add parallel function `file_timer()` that checks the time elapsed since the last FOV file was written. If the elapsed time is greater than 3x the normal FOV timeout period, the watcher function will call the run callbacks and then exit.

**Remaining issues**

- [x] add error check in watcher tests
- [x] add notebook implementations


src/toffy/json_utils.py
```diff
@@ -240,3 +240,39 @@ def check_fov_resolutions(bin_file_dir, run_name, save_path=None):
...
+            os.path.join(bin_file_dir, default_name + ".bin")
+        ) or not os.path.exists(os.path.join(bin_file_dir, default_name + ".json")):
+            missing_fovs[default_name] = [custom_name]
+
+        elif os.path.getsize(os.path.join(bin_file_dir, default_name + ".json")) == 0:
+            missing_fovs[default_name] = [custom_name]
+
+    if missing_fovs:
+        missing_fovs = pd.DataFrame(missing_fovs).to_string(index=False)
+        raise FileNotFoundError(f"The following data was not generated:\n {missing_fovs}")
```
alex-l-kong, line 278: I would personally just warn here and not throw an error.

src/toffy/json_utils.py
```diff
@@ -240,3 +240,39 @@ def check_fov_resolutions(bin_file_dir, run_name, save_path=None):
...
+        default_name = f"fov-{fov_number}-scan-1"
+        custom_name = fov.get("name")
+
+        # add default name and custom name for missing data
+        if not os.path.exists(
+            os.path.join(bin_file_dir, default_name + ".bin")
+        ) or not os.path.exists(os.path.join(bin_file_dir, default_name + ".json")):
+            missing_fovs[default_name] = [custom_name]
+
+        elif os.path.getsize(os.path.join(bin_file_dir, default_name + ".json")) == 0:
```
alex-l-kong, line 273: This will be problematic if both the `.bin` and `.json` file did get written out after the watcher timeout threshold. I think you should check the extracted image folder directory to see if the FOV folder actually exists; it might really be the only check you need.

alex-l-kong (changes requested):
`missing_fov_check` should still error out properly since the watcher also marks FOVs that time out as completed, however, these FOVs won't have any extracted data. The function does need to account for one additional edge case.

src/toffy/json_utils.py
camisowers: This is super helpful, thank you!

camisowers: In the edge case where a FOV times out but actually gets written afterwards, it will not be processed (have an extracted folder). It's probably best to keep the check as is, and warn the user of this processing mishap.

src/toffy/fov_watcher.py
```diff
@@ -198,12 +199,21 @@ def __init__(
...
+            fov_timeout (int):
                 number of seconds to wait for non-null filesize before raising an error
+            watcher_timeout (int):
+                length to wait for new file generation before timing out
         """
         super().__init__()
         self.run_folder = run_folder
 
+        self.last_event_time = datetime.now()
+        self.timer_thread = threading.Thread(
```
srivarra, line 211: I'm confused as to what the thread is up to here. Also does this thread get to be closed eventually?

srivarra (approved):
Looks pretty good, just a clarification question.

src/toffy/fov_watcher.py
```diff
@@ -479,6 +493,40 @@ def on_created(self, event: FileCreatedEvent, check_last_fov: bool = True):
...
+            watcher_timeout (int):
+                length to wait for new file generation before timing out
+        """
+        while True:
+            current_time = datetime.now()
+            time_elapsed = (current_time - self.last_event_time).total_seconds()
+
+            # 3 fov cycles and no new files --> timeout
+            if time_elapsed > watcher_timeout:
+                print("Timed out waiting for new file to be generated.")
```
alex-l-kong, line 511: I'd add, specifically, which FOV it timed out waiting on files for (equivalent to `last_fov_num_updated + 1`).

alex-l-kong (changes requested):
Minor detail about the missing FOV timeout print statement.

alex-l-kong (approved):

ngreenwald (approved):
Looks good

PR #412, opened 2023-08-09
Ensure pulse heights file is named with `_` not `-`

alex-l-kong:

Corrects the way `watcher_callbacks.py` looks for `_pulse_heights.csv` files.

Previously, these pulse height files were specified as `-pulse_heights.csv`. They should be `_pulse_heights.csv` instead.

**Remaining issues**

The MPH pulse files actually do use `-mph_pulse.csv` as a suffix, which is where the confusion arose from. We should think about standardizing this naming scheme.

### DIFF
--- a/src/toffy/watcher_callbacks.py
+++ b/src/toffy/watcher_callbacks.py
@@ -301,7 +301,7 @@ class FovCallbacks:
         if not os.path.exists(pulse_out_dir):
             os.makedirs(pulse_out_dir)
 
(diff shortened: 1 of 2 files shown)


alex-l-kong:
I think I might have pulled a Mariner 1 here guys: https://en.wikipedia.org/wiki/Mariner_1.

alex-l-kong:
@ngreenwald tested that this is now skipping `_pulse_heights.csv` files correctly on your run.

ngreenwald (approved):

PR #417, opened 2023-08-22
Bump to v0.2.1

alex-l-kong:

Adding proficient extraction to the pipeline required an update to `mibi-bin-tools`, a new release will be needed to propagate that.

Version bump.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "Toffy"
-version = "0.0.0"
+version = "0.2.1"
```
srivarra, line 3: The versioning schema is defined by the git tag of the release. Therefore we should not set the version number explicitly in the `pyproject.toml `file as it's generated by that tag during the build and upload to pypi process.

srivarra (changes requested):

pyproject.toml
alex-l-kong: In that case I think we can just close this PR and make a new release off of main?

srivarra: yup!

PR #434, opened 2023-09-29
Updated the `toffy` environment file

srivarra:
**If you haven't already, please read through our [contributing guidelines](https://ark-analysis.readthedocs.io/en/latest/_rtd/contributing.html) before opening your PR**

**What is the purpose of this PR?**

Updates the `toffy` version used in the `environment.yml` files

**How did you implement your changes**

Set the version of `toffy` in the `release-environment.yml` to `0.2.1`.
Created a latest environment which installs whatever version was cloned in the local repo.
Set the default python version for both environments to 3.11.

**Remaining issues**

N/A

PR #436, opened 2023-10-02
Update pulse height generation to be compatible with `mibi-bin-tools==0.2.11`

alex-l-kong:

The new `mibi-bin-tools==0.2.11` update introduces an optimization to `get_median_pulse_height`, allowing the extraction of all specified masses in parallel. The current `toffy` pipeline needs to be updated to reflect that, since currently it calls `get_median_pulse_height` individually per mass.

1. Remove the `for` loop in `write_mph_per_mass` and ensure the pulse heights get returned as a list to `mph_vals` instead.
2. In `compute_mph_metrics`, pass a list containing the single `target` value instead of the `target` value itself to `get_median_pulse_height` call.

**Remaining issues**

Before this gets merged in, #434 should be addressed first.


camisowers (approved):
Nice

srivarra (approved):
Looks Good!

ngreenwald (approved):

PR #443, opened 2023-10-19
Ensure safe conversion of panel types during `modify_panel_ranges`

### LINKED ISSUE #442: Safer dtype handling for `panel_utils.modify_panel_ranges`
**Describe the bug**

The initial `Stop` column in a panel is often a raw integer. The function `modify_panel_ranges` has to add floating point values to this column, which the current version of `pandas` does support. However, future versions of `pandas` will not automatically support this type of conversion.

**Expected behavior**

No warnings thrown right now, and no errors in the future. We'll need to do an explicit cast to fix this.

**To Reproduce**

Run the watcher with a panel file.

alex-l-kong:

Closes #442. A future `pandas` version will error out on `int` to `float` conversions, so we modify that to suppress the warning.

Add the following lines:

```
panel_new["Start"] = panel_new["Start"].astype(np.float64)
panel_new["Stop"] = panel_new["Stop"].astype(np.float64)
```

prior to adding the offsets.

### DIFF
--- a/src/toffy/panel_utils.py
+++ b/src/toffy/panel_utils.py
@@ -129,8 +129,8 @@ def modify_panel_ranges(panel: pd.DataFrame, start_offset: float = 0, stop_offse
     ].index.values
 
     # ensure safe conversion of types
-    panel_new["Start"] = panel_new["Start"].astype(np.float16)
-    panel_new["Stop"] = panel_new["Stop"].astype(np.float16)
+    panel_new["Start"] = panel_new["Start"].astype(np.float64)
+    panel_new["Stop"] = panel_new["Stop"].astype(np.float64)
 
     # add start_offset to 'Start' column
     panel_new.loc[panel_rows_modify, "Start"] = (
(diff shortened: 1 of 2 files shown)


src/toffy/panel_utils.py
```diff
@@ -127,6 +128,10 @@ def modify_panel_ranges(panel: pd.DataFrame, start_offset: float = 0, stop_offse
         (panel_new["Start"] - panel_new["Stop"]).round(1) == -0.3
     ].index.values
 
+    # ensure safe conversion of types
+    panel_new["Start"] = panel_new["Start"].astype(np.float64)
+    panel_new["Stop"] = panel_new["Stop"].astype(np.float64)
```
camisowers, line 133: Does this specifically need to be float64? Could it be float32?

camisowers (approved):
Thanks for putting in the extraction fixes!

srivarra (approved):
Very cool!

PR #444, opened 2023-10-19
Prevent mass column from adding during `rosetta.combine_compensation_files`

alex-l-kong:

If multiple round 2 channel pairs need to be handled, more than 1 round 2 compensation matrix will need to be combined. The compensation values themselves combine correctly, but because the first column (the noisy masses) gets read in as an actual `pandas` column, and not an index, those will also get added together, creating a final noisy mass column that's incorrectly `num_comp_matrices_combine * mass_vals`.

Ensure the first column doesn't get added during this process.

**Remaining issues**

We should strongly consider changing the way we load in Rosetta compensation matrices during the entire process. It's very inconvenient to have the noisy mass values as a separate column and not an index, especially when the masses to denoise are read in as Pandas columns.

### DIFF
--- a/src/toffy/rosetta.py
+++ b/src/toffy/rosetta.py
@@ -188,9 +188,10 @@ def combine_compensation_files(comp_matrix_path, compensation_matrix_names, fina
     )
 
     # loop over the rest and add them in
+    # NOTE: skip the first column, since that just delineates the channels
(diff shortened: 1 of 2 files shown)


srivarra (approved):
LGTM!

camisowers (approved):

PR #455, opened 2023-12-06
Add support for MIBI runs acquired using autogain feature

### LINKED ISSUE #367: Update normalization notebook to work with automatic voltage updates
The current normalization notebook calculates the MPH per image over the course of a run, fits a curve to that decay, and uses that to normalize each channel in each image. Ionpath is going to be rolling out an update to the instrument which automatically adjusts the detector voltage over the course of the run. This will significantly reduce the amount of signal decay we see over the course of the run. 

A side effect of this will be that our current approach for normalizing the data will need to be updated. Specifically, we'll probably want to make a separate notebook, updated_normalization or something like that, since the workflow will be different. Rather than fitting a polynomial, we'll want to check that all of the values are roughly on a straight line. We should then use that fitted flat line as the normalization value for all the FOVs for that particular channel. 

The outlier detection should instead look for values that are far away from that flat line. Some of the other checks (like making sure the voltage in the JSON doesn't change) will need to be tweaked as well.

alex-l-kong:

Closes #367. For autogain runs, the normalization functionality will be different, as the observed MPH values over time do not follow the traditional exponential decay.

Curve fitting should be turned off for autogain. Instead, we default the functionality to what happens if there are fewer than 10 observations in a run: take the median of all MPH points. This will allow for easy visualization and properly adjust the fitted MPH values to account for autogain functionality (better represented by a straight line of fit).

**Remaining issues**

Further testing needed. Additionally, autogain has had ongoing issues with Creed.


tests/normalize_test.py
```diff
@@ -523,10 +524,15 @@ def test_create_fitted_pulse_heights_file(tmpdir, test_zeros, metrics):
...
+                autogain=autogain,
             )
     else:
         df = normalize.create_fitted_pulse_heights_file(
-            pulse_height_dir=pulse_dir, panel_info=panel, norm_dir=tmpdir, mass_obj_func="poly_3"
+            pulse_height_dir=pulse_dir,
+            panel_info=panel,
+            norm_dir=tmpdir,
+            mass_obj_func="poly_3",
+            autogain=autogain,
```
camisowers, line 535: Should we add a test here to make sure `create_fitted_pulse_heights_file()` actually produces a different curve with the autogain flag set to True?

camisowers (changes requested):
One minor comment about adding a test for the new functionality. Also we should remove the warning from the top of notebook 4b that states the notebook is incompatible with the MIBIcontrol 1.8 update.

srivarra (approved):
Looks good to me! Would it be useful for the user to add a link to the [MIBIcontrol v1.8.0 release notes](https://www.ionpath.com/release-notes/)?

bryjcannon:
Looks good. May want to have Rachel test this out on her data if she hasn't already since it's the first fully-autogain-ed cohort and familiar with toffy.

tests/normalize_test.py
alex-l-kong: @camisowers the way I've updated the test is to ensure that a curve that doesn't perfectly fit the observed MPH vals is always fitted without autogain, and the median is always used with autogain. This allows for the testing of different cases if autogain is set or not.

camisowers (approved):

src/toffy/normalize.py
```diff
@@ -707,7 +707,7 @@ def create_fitted_pulse_heights_file(
 
         # for non-autogain runs, use the default min_obs at 10 for curve fitting
         # for autogain runs, the median of all observed MPH values should always be used
-        min_obs = 0 if autogain else 10
+        min_obs = 10**10 if autogain else 10
```
bryjcannon: why did you switch from 0 to 10**10?

alex-l-kong: I got the thresholding wrong. `min_obs` actually defines the minimum number of observations needed to run the curve-fitting process. However, this process should _never_ run for `autogain`. So the hack is to set it to an unreasonably high number that no dataset will ever reach in practice. In this way, the fallback median method is always used.

bryjcannon: Sounds good, thanks!

PR #456, opened 2023-12-13
Prevent pulse_height_fit from throwing dtype warning with floats

alex-l-kong:

When trying to set floating point values to `pulse_height_fit`, the new standard for `pandas` will throw a warning because we initialized them with the _integer_ 0.

Instead of initializing with 0, initialize with 0.0 so floating point values are used.


bryjcannon (approved):
Works - no more dtype warning

PR #458, opened 2024-01-16
Adjust the batch effect QC notebook to average stats across all TMAs

camisowers:

Our current [batch effects notebook](https://github.com/angelolab/toffy/blob/main/templates/6_tma_batch_qc_metrics.ipynb) provides statistics for each TMA grid/control sample. We would like to add an additional plot that compares these metrics across all TMAs to check for any systematic issues. 

1. Change to z-scoring instead of a ranking system, so that we can compare across TMAs. 
2. Remove histogram plots.
3. Average z-score arrays to find averages by RnCm location across all the TMAs in the cohort.
Output:

**Remaining issues**

NA


src/toffy/qc_comp.py
```diff
@@ -651,38 +652,63 @@ def _compute_qc_tma_metrics(self, tma: str):
...
         """
-        with tqdm(total=len(tmas), desc="Computing QC TMA Metric Ranks", unit="TMA") as pbar:
+        max_col, max_row = 0, 0
+        with tqdm(total=len(tmas), desc="Computing QC TMA Metric Z-scores", unit="TMA") as pbar:
             for tma in tmas:
-                self.tma_avg_ranks[tma] = self._compute_qc_tma_metrics_rank(
+                self.tma_avg_zscores[tma] = self._compute_qc_tma_metrics_zscore(
                     tma, channel_exclude=channel_exclude
                 )
+                max_col = (
```
srivarra, line 670: Instead of the `if-else` control flow statement, we can use the built in `max` function which returns the largest item given multiple positional arguments. For example `max_col = max(self.tma_avg_zscores[tma].shape[1], max_col)`.

srivarra (changes requested):
Looks good! Just one suggestion to simplify getting the max row/col.

alex-l-kong (commented):
Looks good otherwise, I'll let @srivarra approve.

srivarra (approved):
Looks good!

ngreenwald (approved):

PR #467, opened 2024-08-22
Enforce `exact_match=True` when listing JSON file for `get_estimated_time` for MPH

### LINKED ISSUE #465: Script 3d sometimes tries to read the wrong json file

**Describe the bug**
Toffy script 3d_compute_median_pulse_height.ipynb first functional code block contains this function:

`mph_comp.compute_mph_metrics(bin_file_dir, mph_out_dir, fov)`

Inside the python code it runs `get_estimated_time(bin_file_dir, fov)` 
-> (inside the above function) `json_file = io_utils.list_files(bin_file_dir, fov + ".json")`
This line finds the first json file that starts with fov name. In my dataset there are 2 json files per fov like that:
Example: fov-1-scan-1.json (the one we want) and fov-1-scan-1_processing.json (the one we get sometimes).
I am uncertain where that second json came from, but it is there for each fov.

**Expected behavior**
The script needs to find the correct json file for each fov. Possible solution: do not invoke `list_files()` function, instead read the correct json directly since we know it's name should be `fov + ".json"`

**To Reproduce**
`TypeError                                 Traceback (most recent call last)
File [~/Documents/Bendall](http://localhost:8888/lab/tree/~/Documents/Bendall) lab projects[/toffy/src/toffy/mph_comp.py:36](http://localhost:8888/toffy/src/toffy/mph_comp.py#line=35), in get_estimated_time(bin_file_dir, fov)
     35     time = run_metadata.get("dwellTimeMillis")
---> 36     estimated_time = int(size**2 * time)
     37 except TypeError:

TypeError: unsupported operand type(s) for ** or pow(): 'NoneType' and 'int'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
Cell In[19], line 9
      7 if not os.path.exists(os.path.join(mph_out_dir, '%s-mph_pulse.csv' % fov)):
      8     print("Computing MPH for fov %s" % fov)
----> 9     mph_comp.compute_mph_metrics(bin_file_dir, mph_out_dir, fov)

File [~/Documents/Bendall](http://localhost:8888/lab/tree/~/Documents/Bendall) lab projects[/toffy/src/toffy/mph_comp.py:108](http://localhost:8888/toffy/src/toffy/mph_comp.py#line=107), in compute_mph_metrics(bin_file_dir, csv_dir, fov, mass, mass_start, mass_stop)
    105     raise FileNotFoundError(f"The FOV name supplied doesn't have a JSON file: {fov}")
    107 count = count_dict[fov]
--> 108 time = get_estimated_time(bin_file_dir, fov)
    110 out_df = pd.DataFrame({"fov": [fov], "MPH": [median], "total_count": [count], "time": [time]})
    112 # saves individual .csv files to csv_dir

File [~/Documents/Bendall](http://localhost:8888/lab/tree/~/Documents/Bendall) lab projects[/toffy/src/toffy/mph_comp.py:38](http://localhost:8888/toffy/src/toffy/mph_comp.py#line=37), in get_estimated_time(bin_file_dir, fov)
     36     estimated_time = int(size**2 * time)
     37 except TypeError:
---> 38     raise KeyError(
     39         "The FOV json file is missing one of the necessary keys (frameSize or dwellTimeMillis)"
     40     )
     42 return estimated_time

KeyError: 'The FOV json file is missing one of the necessary keys (frameSize or dwellTimeMillis)'`

alex-l-kong:

Closes #465. The way `list_files` works is not robust against the new `_processing.json` files that get written to the CACs. For `get_estimated_time`, this causes issues when verifying that a FOV exists in the run folder, then loading it in because sometimes it will attempt to verify the `_processing.json` file and not the main FOV `.json` file.

Make use of `exact_match=True` for `list_files` to enforce hard matching.

**Remaining issues**

Across several places in `toffy` and `ark`, `exact_match=True` must be used, especially because Bryan's ezSegmenter update makes soft matching the default.


alex-l-kong:
The current way `list_files` is implemented for `exact_match=True` is awkward because it removes file extensions when considering matches: https://github.com/angelolab/alpineer/blob/main/src/alpineer/io_utils.py#L78-L83. Worth thinking if a rewrite is necessary.

src/toffy/mph_comp.py
```diff
@@ -24,7 +24,8 @@ def get_estimated_time(bin_file_dir, fov):
     io_utils.validate_paths(bin_file_dir)
 
     # get fov json file in bin_file_path
-    json_file = io_utils.list_files(bin_file_dir, fov + ".json")
+    json_file = sorted(io_utils.list_files(bin_file_dir, fov, exact_match=True))
+    json_file = list(filter(lambda f: ".json" in f, json_file))
```
camisowers, line 28: I think for this case, since we always know we will want the file name `fov + '.json'`, I think we can simply define json_file to be that. It's unnecessary to list the files (returns ['fov-1-scan-1.json', 'fov-1-scan-1.bin']), then just take take the json file only anyways. Then we will ignore the -processing.json without having to specify exact_match.

camisowers (commented):
I agree, the logic for dropping the extension when checking for an exact match is probably a good idea. We would just need to check ark and toffy to make sure that isn't in there for a specific reason. 

I would add a `-processing.json` file to all relevant tests in the pipeline, to see if `list_files()` breaks anywhere else in toffy.

src/toffy/mph_comp.py
alex-l-kong: It won't work if you set the second argument (substr(s)) to `fov-1-scan-1.json` because of the way `list_files` trims the extension for each file listed in `bin_file_dir`, but then compares it to `fov-1-scan-1.json`. So since there's a `.bin` and a `.json` file, you'd be making two comparisons of `fov-1-scan-1` to `fov-1-scan-1.json`, none of which will return True. This caused the previous tests to hang.

srivarra (approved):
Looks good, just making sure you want these print statements there.

tests/fov_watcher_test.py
```diff
@@ -390,32 +406,65 @@ def test_watcher(
                 fovs = fovs[1:]
 
             # extract tiffs check
+            print("TIFF validator check")
```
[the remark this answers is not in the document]
alex-l-kong: @srivarra I think these would be good, previously we didn't know these tests weren't running because of the malformed log file I/O.

camisowers (approved):
Looks good.

PR #470, opened 2024-11-12
Support legacy panel files defined before the Noodle range extraction fix

### LINKED ISSUE #468: `load_panel` should correct `Noodle` range in pre-defined `-toffy.csv` files

Old panels that were extracted with the `-toffy` prefix already will not have the `Noodle` mass range corrected if the current `load_panel` function is used. There should be a workflow to account for this.

A control flow that checks for the correct Noodle range prior to loading the panel in. This correction should also be saved to existing `-toffy` panels.

alex-l-kong:

Closes #468.

Explicitly update the Noodle extraction range on pre-loaded panel files in `load_panel`.

### DIFF
--- a/src/toffy/panel_utils.py
+++ b/src/toffy/panel_utils.py
@@ -274,6 +274,7 @@ def load_panel(panel_path):
                 "'-toffy' from the file name."
             )
         else:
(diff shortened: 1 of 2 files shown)


camisowers (approved):
Looks good!

PR #416, opened 2023-08-17
Check for partially generated fovs

camisowers:
**If you haven't already, please read through our [contributing guidelines](https://ark-analysis.readthedocs.io/en/latest/_rtd/contributing.html) before opening your PR**

**What is the purpose of this PR?**

Closes #400. Checks for incomplete FOV images which are missing data.

**How did you implement your changes**

Since scanning during a run starts at the top of the FOV, if something goes wrong we see this resulting in mostly 0 signal values at the bottom of the image. To be thorough, we check a specified number of images (default 5), and if the amount of non-zero pixels in that last 10 rows is less than 2%, we can identify this as a partial FOV.

**Remaining issues**
N/A

PR #474, opened 2024-11-21
Fix `incomplete_fov_check` to not require a channel named `Au`

### LINKED ISSUE #473: `incomplete_fov_check` errors out if gold mass not specified or not named `Au`
**Describe the bug**

In `bin_extraction.incomplete_fov_check`, the gold channel has two hard requirements as defined at https://github.com/angelolab/toffy/blob/main/src/toffy/bin_extraction.py#L105-L107:

1. It must exist in the panel. If not, then it's assumed the user forgot to add it.
2. It must be named `Au.tiff`.

Each requirement has a problem:
1. Not every cohort extracts gold as a channel. This will especially be the case as more users run MIBIBlue slides.
2. Not everyone names the gold mass in the cohort `Au`.

**Expected behavior**

We should trust that users have specified their panels correctly. The current pipeline extracts every channel that is specified in the panel, so we shouldn't assume `Au` isn't specified in the panel but also was extracted.

Because of the inconsistent gold naming conventions/presences across cohorts, we also need to explicitly specify `extract_intensities` as a `kwarg` or an explicit variable in the extraction notebook for intensity extraction. We can still default it to `["Au", "chan_39"]`, but users will be able to easily modify as needed.

**To Reproduce**

Run a cohort with gold not named `"Au"` or without gold in the panel.

alex-l-kong:

Closes #473.

Remove the addition of the `"Au"` channel in `incomplete_fov_check`.

Add specific lists in the `3a` and `3b` notebooks to allow the user to change their intensity channels they wish to extract (or leave as blank of none specified).

### DIFF
--- a/src/toffy/bin_extraction.py
+++ b/src/toffy/bin_extraction.py
@@ -99,12 +99,14 @@ def incomplete_fov_check(
     run_file_path = os.path.join(bin_file_dir, run_name + ".json")
     run_metadata = read_json_file(run_file_path, encoding="utf-8")
 
-    # get fov and channel info
-    fovs = io_utils.list_folders(extraction_dir, "fov")
+    # TODO: list_folders does not handle cases such as "fov0" correctly
+    # need to add a fix in to alpineer to deal with this
+    fovs = [f for f in os.listdir(extraction_dir) if "fov" in f]
+
+    # NOTE: the specifically-named "Au" channel is no longer be assumed to be present
+    # simply check the first 5 channels found
     channels = io_utils.list_files(os.path.join(extraction_dir, fovs[0]), ".tiff")
     channels_subset = channels[:num_channels]
-    if "Au.tiff" not in channels_subset:
-        channels_subset = channels_subset[:-1] + ["Au.tiff"]
 
     incomplete_fovs = {}
     for fov in fovs:
--- a/templates/3a_monitor_MIBI_run.ipynb
+++ b/templates/3a_monitor_MIBI_run.ipynb
@@ -77,6 +77,17 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# define the specific channels you want to extract as intensity images, leave list empty if there are none\n",
+    "# NOTE: these must match names found in panel_path\n",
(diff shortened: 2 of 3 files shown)


srivarra (approved):

src/toffy/bin_extraction.py
```diff
@@ -99,12 +99,14 @@ def incomplete_fov_check(
...
+    # TODO: list_folders does not handle cases such as "fov0" correctly
+    # need to add a fix in to alpineer to deal with this
+    fovs = [f for f in os.listdir(extraction_dir) if "fov" in f]
+
+    # NOTE: the specifically-named "Au" channel is no longer be assumed to be present
+    # simply check the first "num_channels" channels found
     channels = io_utils.list_files(os.path.join(extraction_dir, fovs[0]), ".tiff")
     channels_subset = channels[:num_channels]
-    if "Au.tiff" not in channels_subset:
-        channels_subset = channels_subset[:-1] + ["Au.tiff"]
```
camisowers, line 107: Noah specifically noted [here](https://github.com/angelolab/toffy/pull/416/files#diff-e6488e9227107ba0a131998f74f9381c2f2a31047d4ad2d7fbc03398a88350a8) that Au should always be a default channel when checking for incomplete fovs. So if some panels don't have gold, then we need this if statement to change instead of removing it completely.
```if "Au.tiff" in channels  and not in channels_subset:```

camisowers (changes requested):
I just have one note that we shouldn't completely remove the Au channel check when looking for incomplete images.

src/toffy/bin_extraction.py
alex-l-kong: @camisowers in line with adding an explicit argument to set `intensity_channels` to the callback creation function, what about another explicit argument defining a custom name for the gold channel? By default, it will be set to `"Au"`, and even if the user leaves it as `None`, it'll still assume an `"Au.tiff"` is found.

In cases where some channels fail to load, instead of erroring out completely, check if the error is specifically caused by a failure to load in gold, and if so, throw a warning that the specific gold channel could not be located and to ask the user to double-check that this is indeed correct.

alex-l-kong:
@jranek need to reopen this issue, no `Au` may be the norm moving forward with MIBIblue. Any thoughts on this?

alex-l-kong:
Addressed by #491

